PR #21, opened 2020-04-21
Pull PSPs into Helm since they are cluster-wide resources

gmalkov:
PSPs were first included in the kustomize YAML to make things simpler in the initial POC. Now that the structure of the cluster-wide Helm chart is more flushed out it makes sense to pull them out here, since currently they are being re-deployed/overwritten on every env deploy.

Related PRs:
https://github.com/broadinstitute/sam/pull/423
https://github.com/DataBiosphere/kernel-service-poc/pull/15
https://github.com/DataBiosphere/terra-workspace-manager/pull/20
https://github.com/DataBiosphere/identity-concentrator/pull/14


charts/terra-cluster-psps/templates/poc.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: kernel-service-poc-service-psp
```
wchamber, line 4: Do we need separate pod security policies for each service? Aren't they the same almost all of the time?

gmalkov: They are for now because we just added boilerplate, very permissive ones for now. We will probably make them more restrictive and different per-app later. This has already started to happen (see the NET_ADMIN permission for Istio). Having PSPs was a security requirement, but I've punted on having them actually restrict anything other than admin access for now.

yonghaoy: At lease for now, seems all services have the same PSP rules and I think most of kernel services will be very similar in the future. We can have only one PSP as "template" and let all apps use this. If some app wants its own, it can include its new psp file in their repo. Otherwise, they will use the template?

wchamber: identity concentrator is more of a special case than most services.
And from Yonghao, did we think other services would also need NET_ADMIN to work with Istio side cars?

gmalkov: Yeah since I wrote my initial comment I realized that is a bad example because it sounds like all the apps need it.
We could definitely make one PSP and have everything use it, and add ones for any services that need special casing in the future. I can make that change and the corresponding changes in the service PRs.
I have no idea really how much we will change the PSPs per-service in the future so I don't have a strong opinion on any particular way to manage them. The single PSP definitely seems neater for now. Stand by for updates to all the PRs.

wchamber: Does the same logic also apply to Roles?

gmalkov: It could if we switch to clusterRoles and clusterRoleBindings, then we could define them once per app cluster-wide (I don't think we should have a single role for all apps, there is all sorts of RBAC stuff I'd like to have the flexibility to do per-app)

gmalkov: Roles can control a lot more than just PSP use though, so I am a little hesitant to make this switch as I can see us potentially making some environments (prod) more restrictive than others 🤔

PR #29, opened 2020-05-08
Add Sam, OpenDJ charts

gmalkov:


charts/opendj/templates/stateful-set.yaml
```diff
@@ -0,0 +1,57 @@
...
+          protocol: TCP
+        - containerPort: 4444
+          protocol: TCP
+        resources: {}
+        terminationMessagePath: /dev/termination-log
+        terminationMessagePolicy: File
+        env:
+        - name: BASE_DN
+          # This value should be replaced by environment.
+          value: "dc=dsde-my-environment,dc=broadinstitute,dc=org"
```
gmalkov, line 41:
```suggestion
          value: "dc=dsde-my-environment,dc=broadinstitute,dc=org"
```
Not familiar enough with this to know the proper syntax here. What are the valid options? If we set it to say `dc=dsde-{{ .Values.global.terraEnv }}` will it only work for current Terra environments?

wchamber: This is one of those values that spans a few places. It's the openDJ base domain name, so all the things that talk to it need to be using the same value. I don't know if the broadinstitute org values are special.

firecloud develop definition: https://github.com/broadinstitute/firecloud-develop/blob/0bd6665585421bb8302799e10adb810e67ebef74/configure.rb#L56

Sam today: https://github.com/broadinstitute/firecloud-develop/blob/dev/base-configs/sam/sam.conf.ctmpl#L32
and every Apache proxy configuration (which we will need in some form but don't have yet for k8s.): https://github.com/broadinstitute/firecloud-develop/search?q=LDAP_BASE_DOMAIN&unscoped_q=LDAP_BASE_DOMAIN

charts/sam/templates/deployment.yaml
```diff
@@ -0,0 +1,67 @@
...
+                -Xms8g -Xmx8g -XX:NewSize=1g -XX:MaxNewSize=1g -XX:SurvivorRatio=8 -XX:+UseParNewGC
+                -XX:+UseConcMarkSweepGC -XX:+CMSParallelRemarkEnabled -XX:+UseCMSInitiatingOccupancyOnly
+                -XX:CMSInitiatingOccupancyFraction=60 -XX:+ScavengeBeforeFullGC -XX:+CMSScavengeBeforeRemark
+                -Xloggc:/var/log/gc/sam-gc.log -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=2 -XX:GCLogFileSize=100M
+                -XX:+PrintGCDateStamps -verbose:gc -XX:+PrintGCDetails -XX:+PrintClassHistogramBeforeFullGC
+                -XX:+PrintClassHistogramAfterFullGC -Dconfig.file=/etc/sam-configs/sam.conf
+                -Dcom.sun.jndi.ldap.connect.pool.protocol=ssl -Dcom.sun.jndi.ldap.connect.pool.maxsize=100
+          volumeMounts:
+            - name: sam-configs
+              # TODO need to change rendered sam.conf references to add sam-configs subdir.
```
wchamber, line 40: oops, please remove this TODO, I did it and forgot to remove.

charts/sam/templates/deployment.yaml
```diff
@@ -0,0 +1,67 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: sam-deployment
+  {{- template "sam.labels" }}
+spec:
+  # TODO consider changing the replicas number or always overlaying it for different environments.
```
wchamber, line 7: doesn't have to be this PR, but this is an example of something we would add a chart variable for, right?

charts/opendj/templates/stateful-set.yaml
wchamber: but to answer your question, there's nothing special about the 'dsde-env' value AFAIK, we just need to agree on it.

charts/sam/templates/deployment.yaml
gmalkov: Yep, I'll add it. Mostly I just copy/pasted stuff and got rid of the kustomize-specific bits

charts/opendj/templates/stateful-set.yaml
gmalkov: OK, I'll make it dsde-[env] so that it's easy to keep it consistent with the current configs

gmalkov:
> Is there a definition of the inputs to these charts somewhere? Or not needed cause they use default chart values?

Yeah the values definitions are going to be added to the terra-helmfile repo in another PR. That's going to be the repo for all of the Terra env values overrides.

charts/sam/values.yaml
```diff
@@ -0,0 +1,9 @@
+image:
+  repository: gcr.io/broad-dsp-gcr-public/sam
+  # Override the image tag to deploy by setting this variable.
+  # If no value is set, the chart's appVersion will be used.
+  # tag: latest
+gateway:
```
wchamber, line 6: worth a comment on what this controls, or is the style to go and lookup where it's used?

charts/opendj/templates/stateful-set.yaml
```diff
@@ -0,0 +1,57 @@
...
+          protocol: TCP
+        - containerPort: 4444
+          protocol: TCP
+        resources: {}
+        terminationMessagePath: /dev/termination-log
+        terminationMessagePolicy: File
+        env:
+        - name: BASE_DN
+          # This value should be replaced by environment.
+          value: "dc=dsde-{{ .Values.global.terraEnv }},dc=broadinstitute,dc=org"
```
wchamber, line 41: So this gets defined in another values file? Is there any way to trace that file to the fact that opendj uses it here? Feels a little buried and odd partially having its own values file and partially relying on the terra global values.

I think we might even want to make the whole base domain value a template value since the whole string is necessary, it's just happens to be the case only the env part is modified.

charts/sam/templates/deployment.yaml
wchamber: remove the comment too.

charts/opendj/templates/stateful-set.yaml
```diff
@@ -0,0 +1,57 @@
...
+        - containerPort: 1636
+          protocol: TCP
+        - containerPort: 4444
+          protocol: TCP
+        resources: {}
+        terminationMessagePath: /dev/termination-log
+        terminationMessagePolicy: File
+        env:
+        - name: BASE_DN
+          # This value should be replaced by environment.
```
wchamber, line 40: remove comment

charts/opendj/templates/stateful-set.yaml
gmalkov: Makes sense, I'll make it a values file parameter

charts/sam/values.yaml
gmalkov: Sure

PR #32, opened 2020-05-14
Add workspace manager chart

gmalkov:


charts/workspacemanager/values.yaml
```diff
@@ -0,0 +1,21 @@
...
+  # Override the image tag to deploy by setting this variable.
+  # If no value is set, the chart's appVersion value will be used.
+  # tag: latest
+  imagePullPolicy: Always
+vault:
+  enabled: true
+  # pathPrefix is required if Vault secret sync is enabled
+  # pathPrefix: secret/dsde/terra/kernel/integration
+
+samAddress: https://sam.dsde-dev.broadinstitute.org/
```
zloery, line 18: Nothing to do in this PR, but eventually we should change how services get addresses. That probably changes after we get Sam on K8s, though.

PR #37, opened 2020-05-26
add istio config for authN and authZ

yonghaoy:
Tested locally got expect result.

AuthN is moved from https://github.com/DataBiosphere/env-base/pull/1


charts/istioconfig/Chart.yaml
```diff
@@ -0,0 +1,22 @@
+---
+apiVersion: v2
+name: terra-istio-config
```
gmalkov, line 3: The name of the chart should be the name of the folder. `istioconfig` is better since we don't want dashes in our chart names. See naming doc below:
https://github.com/broadinstitute/terra-helmfile/blob/master/docs/NAMING.md

charts/istioconfig/values.yaml
```diff
@@ -0,0 +1,5 @@
+namespaces:
+  istio: istio-system
+
+gateway:
+  name: template-gateway
```
gmalkov, line 5: I don't think either of these values are ever used in the chart, why are they here?

charts/istioconfig/templates/service-authorization.yaml
```diff
@@ -0,0 +1,15 @@
+# Sample AuthZ Policy for future istio.
+# Block traffic for kernel-service-poc
+apiVersion: security.istio.io/v1beta1
+kind: AuthorizationPolicy
+metadata:
+  name: "poc-access"
+  namespace: terra-yyu
```
gmalkov, line 7: You do not need to specify a namespace, since this chart will be applied in an environment namespace by ArgoCD and the namespace will be automatically populated.

charts/istioconfig/templates/request-header-checker.yaml
```diff
@@ -0,0 +1,24 @@
+# Define a rule to reject all request if oidc_access_token is presented in request header.
```
gmalkov, line 1: Are the configurations in this file specific to a namespace or global? Any configuration not specific to a namespace/environment, should instead go in the terra-cluster-networking chart.

charts/istioconfig/templates/jwt-checker.yaml
```diff
@@ -0,0 +1,14 @@
+# policy for JWT Verification. All services listed in targets field would require JWT token
+apiVersion: "authentication.istio.io/v1alpha1"
+kind: "Policy"
+metadata:
+  name: "jwt-example"
+spec:
+  targets:
+    - name: kernel-service-poc-service
```
gmalkov, line 8: This list of targets is a good candidate for pulling out into a values. We can have a value that's a list of service names and then loop over them here.
https://helm.sh/docs/chart_template_guide/control_structures/#looping-with-the-range-action

charts/istioconfig/templates/service-authorization.yaml
yonghaoy: ah, that is used for my local testing. deleted.

charts/istioconfig/templates/request-header-checker.yaml
yonghaoy: resolved offline

charts/istioconfig/templates/service-authorization.yaml
```diff
@@ -0,0 +1,18 @@
+# Sample AuthZ Policy for future istio.
```
gmalkov, line 1: What do you mean by future Istio here? 1.4 that we just upgraded to or something later?

charts/istioconfig/Chart.yaml
```diff
@@ -0,0 +1,22 @@
+---
+apiVersion: v2
+name: istioconfig
+description: A Helm chart to steup istio configs
```
gmalkov, line 4:
```suggestion
description: A Helm chart to setup Istio configs
```

charts/istioconfig/templates/service-authorization.yaml
yonghaoy: ahh, sorry, forgot to change that

PR #44, opened 2020-06-03
Add OIDC proxy image to WM deployment

zloery:
Adds OIDC proxy in front of WM, with configurations to send info to the proxy instead of directly to WM if it's being used. The proxy terminates SSL connections and should be configured to work with tcell, though it isn't yet.


charts/workspacemanager/templates/proxy-configmap.yaml
```diff
@@ -0,0 +1,78 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: terra-workspace-manager-proxy-configmap
+  labels:
+    {{ if not .Values.devDeploy }}{{ template "workspacemanager.labels" }}{{ end }}
+data:
+  siteconf: |-
+    ServerAdmin ${SERVER_ADMIN}
+    ServerName {{ .Chart.Name }}.{{ .Release.Namespace }}.{{ .Values.certManager.dnsSuffix }}
```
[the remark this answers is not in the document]
gmalkov: Ah yep, `dnsSuffix` should not be under `certManager` really. For existing Terra envs it will be set to `dsde-[env].broadinstitute.org` and instead of cert-manager creating certs for each app we will just sync the existing wildcard certs for the corresponding env from Vault. I can re-factor this real quick.

charts/workspacemanager/templates/deployment.yaml
```diff
@@ -87,7 +89,37 @@ spec:
         - name: cloudsql-sa-creds
           mountPath: /secrets/cloudsql
           readOnly: true
+      {{- if .Values.enableProxy }}
+      - name: oidc-proxy
+        image: "{{ .Values.proxyImage.repository }}:{{ .Values.proxyImage.version }}"
+        ports:
+        - containerPort: 443
+        - containerPort: 80
```
gmalkov, line 97: Is there a reason to even expose port 80 when we have the proxy set up? Don't we want all traffic to be https then?

zloery: I put that in because it seemed nicer to leave 80 open and having the proxy redirect to https instead of just timing out connections to 80. Not a strong reason, so happy to change this if that's a bad practice

rmeffan: Listening on port 80 is a best practice, as long as you send an appropriate redirect to https. @zloery's response is spot-on:

> I put that in because it seemed nicer to leave 80 open and have the proxy redirect to https instead of just timing out connections to 80.

ETA: This can be handled upstream in the connection chain (at an LB or other ingress point), but it's good behavior to handle common errors like something using http in a URL.

gmalkov: I see now, got it!

charts/workspacemanager/templates/deployment.yaml
```diff
@@ -87,7 +89,37 @@ spec:
...
+        volumeMounts:
+        - mountPath: /etc/apache2/sites-available/site.conf
```
rmeffan, line 99: Recommend using a descriptive name vs. the generic `site.conf` here. The sites-available and `sites-enabled` pattern allows for `0..N` entries in these directories. However, it's still better for human debugging and potential automation uses to use a standard naming pattern prefix (or name) for these types of conf files.

charts/workspacemanager/templates/deployment.yaml
```diff
@@ -87,7 +89,37 @@ spec:
...
+          name: siteconf
+          subPath: site.conf
+        - mountPath: /etc/ssl/certs/server.crt
```
rmeffan, line 102: This is a standard pattern that we use in many places. The only issue with it is that it assumes a single TLS cert key in a given location. The problem with this is that /etc/ssl/certs contain more than one cert, and /etc/ssl/private can contain more than one key.

Notes:
1. If there will only be one TLS cert and key in this location for all of this type of use case, this is not a required change.
2. Changing this naming pattern may require a rethink about how we name and provision TLS artifacts. Therefore, assuming that Note 1 is the case, this can be deferred.

charts/workspacemanager/templates/proxy-configmap.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+          ${AUTH_TYPE2}
+          ${AUTH_REQUIRE2}
+
+          ProxyPass http://localhost:8080/api
+          ProxyPassReverse http://localhost:8080/api
+
+          AddOutputFilterByType DEFLATE application/json text/plain text/html application/javascript application/x-javascript
+        </Location>
+    </VirtualHost>
```
rmeffan, line 78: Missing newline; you should add one.

charts/workspacemanager/templates/proxy-configmap.yaml
```diff
@@ -0,0 +1,78 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: terra-workspace-manager-proxy-configmap
+  labels:
+    {{ if not .Values.devDeploy }}{{ template "workspacemanager.labels" }}{{ end }}
+data:
+  siteconf: |-
```
rmeffan, line 8: Related somewhat to the comment earlier about site.conf being the filename, would it be beneficial for other people who encounter this file if it had the following name: `terra-workspace-manager-proxy-apache-httpd-configmap`?

charts/workspacemanager/templates/proxy-configmap.yaml
```diff
@@ -0,0 +1,78 @@
+apiVersion: v1
```
rmeffan, line 1: General comment about this file: Do you need me to do a close review of this file? If so, it would be helpful if I could view a rendered copy of it.

That said, it looks correct based on a quick pass through it.

charts/workspacemanager/templates/deployment.yaml
zloery: That's actually good to know - I don't think we'd ever use more than one entry. Should we be using a different pattern (Google points me to using `httpd.conf`)? Just renaming this one in place seemed to cause issues with the proxy not reading the config.

charts/workspacemanager/templates/deployment.yaml
zloery: Note 1 is true, there should only be one TLS cert + key for the server

charts/workspacemanager/templates/proxy-configmap.yaml
zloery: Renamed to `proxy-apache-httpd-configmap`. Adding workspace manager feels redundant given the directory structure.

charts/workspacemanager/templates/proxy-configmap.yaml
zloery: A quick pass is probably the right level of review for now, especially since I still need to come back and add in tcell configuration at some point.
Thank you for the comments!

PR #45, opened 2020-06-03
Mf cert manager

mflinn-broad:
Helm chart which installs cert manager using the jetstack/cert-manager as a base and then configures letsencrypt cas able to issue certificate under the *.envs.broadinstitute.org and *.dsp-devops.broadinstitute.org domains. 

The current version of this chart installs cert-manager v0.10.1 which is deprecated. This is a workaround for an issue where the GKE installation of istio forces the cluster to install CRDs for a deprecated version of istio. They cannot be deleted and cause conflicts if they exist simultaneously with the current cert-manager CRDs.  v0.10.1 is the most modern version of cert-manager that uses the deprecated CRDS enforced by GKE istio. 

Once we are managing our own istio deployment this chart will be revised to use a newer version of cert-manager.


smark88:
I might be late to the party but have you looked into Google managed cert s

https://cloud.google.com/kubernetes-engine/docs/how-to/managed-certs

I added it into my oidc chart. 

https://github.com/broadinstitute/datarepo-helm/blob/master/charts/oidc-proxy/templates/gcpmanagedcert.yaml
https://github.com/broadinstitute/datarepo-helm/blob/3736dc7f98d4a0490c8c6e069bd12b8a5210b56d/charts/oidc-proxy/templates/oidc-ingress.yaml#L10

mflinn-broad:
@smark88 Yes I did look into google managed certs. The main concerns were lack of flexibility, (let's encrypt only, no wildcards?), and concern over being too dependent on google managed services when moving to other cloud platforms.

charts/cert-manager/templates/clusterrole.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata: 
+  name: {{ .Release.Name }}-ClusterRole
+  namespace: {{ .Release.Namespace }}
```
gmalkov, line 5: Helm should not define a namespace
https://github.com/helm/helm/issues/5465

charts/cert-manager/values.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+# Email that will be used to register issuer with lets encrypt
+# Will receive communications from lets encrypt here
+issuerAccountEmail: ap-devops@broadinstitute.org
+dnsProject:
+  # Id for the GCP project that will host your cloud DNS resources
+  id: dsp-devops
+  serviceAccount:
+    # name of the k8s secret that will hold projects SA credentials 
+    secretName: something
```
gmalkov, line 18: A bunch of these values really shouldn't have defaults, but should always be required to be overwritten in values files. The pattern I've been using for such values is:
- Comment out the value in the chart
- Add a [required block](https://helm.sh/docs/howto/charts_tips_and_tricks/#know-your-template-functions) around anywhere it is used so that there's a helpful error message if it's omitted.

mflinn-broad: ty for the suggestions! will make these changes.

smark88:
> @smark88 Yes I did look into google managed certs. The main concerns were lack of flexibility, (let's encrypt only, no wildcards?), and concern over being too dependent on google managed services when moving to other cloud platforms.

In a sense it more secure, but also SANS are now supported so you could have a SANs for other fqdns but i see the ease of the wildcard for a larger env.

PR #47, opened 2020-06-08
Add cert, make service a k8s LB

gmalkov:


charts/workspacemanager/templates/cert.yaml
```diff
@@ -0,0 +1,16 @@
+{{ if .Values.certManager.enabled -}}
```
zloery, line 1: Style question: why `-}}` over `{{-`?

gmalkov: Really no reason, I can change it

PR #54, opened 2020-06-15
Add a way to pull certs from Vault

gmalkov:


charts/workspacemanager/templates/cert.yaml
```diff
@@ -6,11 +6,27 @@ metadata:
...
+spec:
+  name: {{ .Chart.Name }}-cert
+  keysMap:
+    tls.crt:
+      path: {{ required "A valid vaultCert.cert.path value is required when vaultCert is enabled" .Values.vaultCert.cert.path }}
+      key: {{ required "A valid vaultCert.cert.secretKey value is required when vaultCert is enabled" .Values.vaultCert.cert.secretKey }}
+    tls.key:
+      path: {{ required "A valid vaultCert.key.path value is required when vaultCert is enabled" .Values.vaultCert.key.path }}
+      key: {{ required "A valid vaultCert.key.secretKey value is required when vaultCert is enabled" .Values.vaultCert.key.secretKey }}
 {{ end -}}
```
mflinn-broad, line 32: We are going to need to add keys for the intermediate certificate to this secret also. Then mount them to the apache proxy. They will likely need to be consumed in the same manner used by the legacy firecloud apache proxies

mflinn-broad: The above is in order to use the old wildcard certificates from vault

gmalkov: Done, thanks for pointing this out

PR #62, opened 2020-07-01
CRL Janitor Update

gmalkov:
- Added helm-docs annotations and generated README
- Added apache proxy for TLS


gmalkov:
> Just curious, why don't we have a sub-chart for all the apache proxy, firewall & cert stuff? Seems like it'll be pretty much the same for many services.

Yeah for sure, we are definitely getting to the point where it makes sense to pull out a bunch of common pieces into [https://helm.sh/docs/topics/library_charts/](library charts). We have discussed this, just haven't gotten around to it yet.

PR #71, opened 2020-07-14
DDO-512 Implement JSON-style logging for Cromwell with a custom logback.xml file

choover-broad:
Configure cromwell on GKE to use a custom logback.xml that produces single-line JSON log messages.

This is NOT real structured logging, where messages are serialized by the application into JSON. I think that would require an application-side change to, at minimum, pull in libraries from [logback-contrib](https://github.com/qos-ch/logback-contrib/wiki/JSON).

Instead, a Pattern is used to to format the message into a JSON object. Logback's `replace` function is used to escape newlines and double quotes in messages, as well as newlines in exception stacktraces.

### DIFF
--- a/charts/cromwell/Chart.yaml
+++ b/charts/cromwell/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: cromwell
-version: 0.5.1
+version: 0.6.0
 type: application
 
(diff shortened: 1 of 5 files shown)


choover-broad:
cc @aednichols in case this is of interest!

PR #72, opened 2020-07-14
AS-367: Add service google project and sampling probability for workspace manager; add new trace SA 

ansingh7115:


zloery:
Adding Greg to this - I forget if GKE exposes google project already somehow, but he might know. If not, LGTM

PR #73, opened 2020-07-15
DDO-398-prometheus-argocd

mflinn-broad:
We are planning on deploying prometheus in argocd as a dependency of the cluster charts. This pr facilitates this by enabling
prometheus to still be installed in the monitoring namespace when used as dependency when the parent is installed in a different namespace.


mflinn-broad:
> LGTM! this is a matter of taste, but should `argoCd.namespaceOverride` just be `namespaceOverride`? Matches the `prometheus-operator` convention, and the feature itself isn't argocd-specific... Up to you!

My thinking was that this value would only be used when deploying with argo, but that may or may not prove to be true down the road. I'll take it out for now, looks cleaner anyways

PR #82, opened 2020-07-20
DDO-544 Cromwell chart improvements

choover-broad:
* Escape tabs in log messages (exception stacktraces have them)
* Automatically restart Cromwell when configmap is changed
* Use helm-docs to generate README.md

### DIFF
--- a/charts/cromwell/templates/_configmap.tpl
+++ b/charts/cromwell/templates/_configmap.tpl
@@ -1,13 +1,30 @@
+{{- /* Generate the data component of a Cromwell config map */ -}}
+{{- define "cromwell.configmap.data" -}}
+data:
+  logback.xml: |
+{{ include "cromwell.config.logback" . | indent 4 }}
+{{- end -}}
+
 {{- /* Generate a configmap for a Cromwell deployment */ -}}
 {{- define "cromwell.configmap" -}}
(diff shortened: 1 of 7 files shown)


charts/cromwell/templates/_configmap.tpl
```diff
@@ -1,13 +1,30 @@
...
 {{- define "cromwell.configmap" -}}
 {{- $settings := ._deploymentSettings -}}
+
+{{- /*
+  Render ConfigMap data, then checksum it, and store the checksum
+  in the deploymentOutputs dictionary so that it be included as an annotation
+  on the deployment's pod template. (see _deployment.tpl)
+
+  This is used to automatically restart Cromwell pods when the ConfigMap
+  changes.
```
mflinn-broad, line 18: Is the idea here that when the config map is updated, the checksum annotation will be updated on the deployment. This is enough to cause the controller to think the deployment has changed and will automatically perform a redeploy? If so awesome!

choover-broad: Yep! Read about this technique in a random blog post somewhere, lost the link now. I definitely prefer this approach for configmap changes where contents of the config map are all deterministic / not updated by another service!

PR #86, opened 2020-07-23
DDO-548 Add support for ephemeral postgres container

gmalkov:
[Postgres library chart PR](https://github.com/broadinstitute/terra-helm/pull/85)

### DIFF
--- a/charts/poc/templates/deployment.yaml
+++ b/charts/poc/templates/deployment.yaml
@@ -27,23 +27,36 @@ spec:
         {{- if not .Values.proxy.enabled }}
         ports:
         - containerPort: 8080
-        {{- end}}
+        {{- end }}
         env:
         - name: DATABASE_USER
+          {{- if not .Values.postgres.enabled }}
           valueFrom:
             secretKeyRef:
               name: {{ .Chart.Name }}-postgres-db-creds
               key: username
+          {{- else }}
+          value: postgres
+          {{- end }}
         - name: DATABASE_USER_PASSWORD
+          {{- if not .Values.postgres.enabled }}
           valueFrom:
             secretKeyRef:
               name: {{ .Chart.Name }}-postgres-db-creds
               key: password
+          {{- else }}
+          value: {{ .Values.postgres.password }}
+          {{- end }}
         - name: DATABASE_NAME
+          {{- if not .Values.postgres.enabled }}
           valueFrom:
             secretKeyRef:
               name: {{ .Chart.Name }}-postgres-db-creds
               key: db
+          {{- else }}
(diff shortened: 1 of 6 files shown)


charts/poc/templates/deployment.yaml
```diff
@@ -72,6 +85,9 @@ spec:
         - name: sa-creds
           mountPath: /secrets/sa
           readOnly: true
+      {{- else }}
+      {{- include "postgres.container.tpl" . | indent 6 }}
```
choover-broad, line 89: Hmm, it's interesting that Postgres is deployed as a container in the pod! I think it could be okay as long as there's only ever 1 replica, and we don't care about starting with an empty DB on every app re-deploy...

How will idempotent re-deploys on the PR integration test workflow work with this?
* Environment is created and all apps start with fresh databases, populating schemas on startup
* Integration tests run, and populate some data as they interact with several services
* PR branch is updated, and the app under test is re-deployed, gets a fresh database. Other apps are using the old database... Is this going to lead to unexpected behavior? Like, where the Rawls database contains a reference to a workflow that does not exist in the Cromwell database?

I think it would probably be okay in most cases, but I think it could lead to fewer surprises down the line if the Postgres instance is a separate Deployment. I can definitely imagine an engineer adjusting the number of app replicas for testing purposes and getting burned.

gmalkov: As per our [discussion on preview environments](https://docs.google.com/document/d/1lTIZQc5UMRQfZTsSCaCFYHrzw8U--zf9nC76MBHv764/edit?ts=5f17576f#heading=h.pmnouop8f0nb), the initial use case assumes an environment is dirty after a test run and will always be nuked and re-created. I made this implementation with that use case in mind.

The replica thing is a good point and something I didn't think of. I can definitely make this its own deployment that can be configured with as many dbs as needed for all of the apps. Can use a cluster IP and have all apps in the env talk to the same deployment. That should be pretty easy to do. Thoughts?

choover-broad: What about having each app declare a conditional dependency on a shared postgres chart, as you've done with the library chart?
* If ArgoCD is used for these environments (IDK if it will be), the Postgres instance will appear in the application's Argo app
* Different apps can use different versions of Postgres

Open to a shared instance if it's easier, I think it means having to maintain a separate set of values whenever we add or remove an app, though

gmalkov: Yep sure that does sound best, still using a library chart but a separate deployment. Will do. Clearly I should have thought this through more. Have been dealing with a lot of context switching so wasn't able to focus on this as much as I'd have liked.

PR #94, opened 2020-07-30
[PF-61] Add pubsub configs for crl-janitor

yonghaoy:


charts/crljanitor/templates/deployment.yaml
```diff
@@ -59,6 +59,21 @@ spec:
             secretKeyRef:
               name: crl-janitor-stairway-db-creds
               key: db
+        - name: PUBSUB_PROJECT_ID
```
wchamber, line 62: wdyt about identifying what the pubsub project/topic/subscription are used for in the janitor? Is it worth saying that this is for the track resource input? Or is there only ever going to be one unqualified pubsub thing so that it's obvious?

yonghaoy: add TRACK_RESOURCE_ as prefix. PTAL. Thanks

charts/crljanitor/templates/secrets.yaml
```diff
@@ -69,4 +69,23 @@ spec:
...
       key: key
+---
+apiVersion: secrets-manager.tuenti.io/v1alpha1
+kind: SecretDefinition
+metadata:
+  name: secretdefinition-crl-janitor-pubsub
+  labels:
+    {{ template "crljanitor.labels" . }}
+spec:
+  name: crl-janitor-pubsub
```
wchamber, line 80: crl-janitor-track-resource-pubsub?
ditto metadata.name

yonghaoy: done

PR #98, opened 2020-08-04
ontology first pass

rushtong:
Ontology k8s set up for Terra
See also: https://github.com/broadinstitute/terra-helmfile/pull/81


rushtong:
Remove dashes - use all one word lc.

choover-broad:
TODO: rename templates and labels to `ontology` instead of `consent-ontology`

choover-broad:
TODO: Update this chart to support the value `global.applicationVersion`, eg: https://github.com/broadinstitute/terra-helm/blob/master/charts/cromwell/values.yaml#L2

All Terra app helm charts should conventionally support this value. It tells the chart what version of the application to deploy, and should be substituted into image tags, labels, configmaps, and so on wherever it is needed.

charts/ontology/templates/service.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: {{ .Chart.Name }}
+  labels:
+{{ include "ontology.labels" . | indent 4 }}
+spec:
+  ports:
+    - name: http
+      port: 8080
+  selector:
+    app: {{ .Chart.Name }}-service
```
mflinn-broad, line 12:
```suggestion
  selector:
    app: {{ .Chart.Name }}
```
This field must be used to select a label that is present on ontology pods. 
Becasue all ontology pods just have the label `app: ontology` this is currently an empty service that doesn't select any pods.

charts/ontology/templates/service.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ .Chart.Name }}
+  labels:
```
mflinn-broad, line 5:
```suggestion
  labels:
    app: {{ .Chart.Name }}-service
```
If you want to include a label to indicate that this is the ontology service. It should be added to the labels on the service itself rather than be in the seletor

charts/ontology/templates/deployment.yaml
```diff
@@ -0,0 +1,78 @@
...
+{{ include "ontology.labels" . | indent 4 }}
+spec:
+  replicas: {{ .Values.replicas }}
+  selector:
+    matchLabels:
+      app: {{ .Chart.Name }}
+  template:
+    metadata:
+      name: {{ .Chart.Name }}
+      labels:
```
mflinn-broad, line 15:
```suggestion
      labels:
{{ include "ontology.labels" . | indent 8 }}
```
It's a good practice to have the chart template labels on the pods as well so that they can be used for easier selection/searching

charts/ontology/templates/service.yaml
```diff
@@ -0,0 +1,12 @@
...
+{{ include "ontology.labels" . | indent 4 }}
+spec:
+  ports:
+    - name: http
+      port: 8080
+  selector:
```
mflinn-broad, line 11:
```suggestion
  type: LoadBalancer
  selector:
```

This will enable the service to be accessible outside the cluster. For now this will randomnly generate a new ip address each time the service is created. I think we can worry about static ip and dns once we confirm the service is functioning

charts/ontology/templates/service.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ .Chart.Name }}
+  labels:
+{{ include "ontology.labels" . | indent 4 }}
+spec:
+  ports:
+    - name: http
+      port: 8080
```
mflinn-broad, line 10:
```suggestion
  ports:
    - name: http
      protocol: TCP
      port: 80
    - name: https
      protocol: TCP
      port: 443
```
Now that the proxy is up and running I think the service should communicate with those ports rather than directly with the ontology container

mflinn-broad: nevermind, I misread the chart on this one.

charts/ontology/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+
+
+
+## Chart Values
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| global.applicationVersion | string | Is set by Helmfile on deploy | Ontology global version |
+| appVersion | string | Is set by Helmfile on deploy | Ontology image version/tag. Required unless using `image`. |
```
choover-broad, line 16: Please use `global.applicationVersion` instead of `appVersion`. (Sorry, I know this is confusing and not documented anywhere...)

Our initial terra-helmfile implementation supplied a value called `appVersion` to all app charts but we ended up changing that to a global value so that it could be consumed by subcharts. For now, terra-helmfile supplies both values, but `global.applicationVersion `is the correct value to use, `appVersion` is deprecated. All Terra app charts that have a value called `appVersion` should rename the value to `global.applicationVersion`.

charts/ontology/README.md
```diff
@@ -0,0 +1,79 @@
...
+| devDeploy | bool | `false` | Enable to deploy to dev locally with Skaffold |
+| elasticSearch.server1 | string | `nil` | first elastic search host |
+| elasticSearch.server2 | string | `nil` | second elastic search host |
+| elasticSearch.server3 | string | `nil` | third elastic search host |
+| environment | string | `nil` | Environment, [dev, staging, prod] |
+| gcsAccount.path | string | `nil` | Vault path to GCS base64 service account json |
```
choover-broad, line 22: There are often good reasons to break this rule (have done it many times myself and lots of our charts have nested values), but just as an FYI, flat values are preferred to nested ones in official Helm best practices: https://helm.sh/docs/chart_best_practices/values/#flat-or-nested-values

charts/ontology/templates/deployment.yaml
```diff
@@ -0,0 +1,94 @@
...
+      hostAliases:
+        - ip: 127.0.0.1
+          hostnames:
+            - app
+      containers:
+        - name: {{ .Chart.Name }}-service
+          image: "{{- if .Values.image -}}
+          {{ .Values.image }}
+          {{- else -}}
+          {{ .Values.imageConfig.repository }}:{{ default .Values.appVersion .Values.imageConfig.tag }}
```
choover-broad, line 32: Use `global.applicationVersion` instead of appVersion as mentioned above

charts/ontology/README.md
```diff
@@ -0,0 +1,79 @@
...
+| vault.pathPrefix | string | `nil` | Vault path prefix for secrets. Required if vault.enabled. |
+| vaultCert.enabled | bool | `false` | Enable to sync certificate secret from Vault with secrets-manager |
+| vaultCert.cert.path | string | `nil` | Path to secret containing .crt |
+| vaultCert.cert.secretKey | string | `nil` | Key in secret containing .crt |
+| vaultCert.key.path | string | `nil` | Path to secret containing .key |
+| vaultCert.key.secretKey | string | `nil` | Key in secret containing .key |
+| vaultCert.chain.path | string | `nil` | Path to secret containing intermediate .crt |
+| vaultCert.chain.secretKey | string | `nil` | Key in secret containing intermediate .crt |
+
+## Cheat Sheet
```
choover-broad, line 43: This documentation is excellent!

charts/ontology/values.yaml
```diff
@@ -0,0 +1,68 @@
+# image is used for local Skaffold deploys
+global:
+  # global.applicationVersion -- What version of the Cromwell application to deploy
+  applicationVersion: latest
+
+appVersion: 1
```
choover-broad, line 6: Remove, as mentioned

charts/ontology/values.yaml
```diff
@@ -0,0 +1,68 @@
+# image is used for local Skaffold deploys
+global:
+  # global.applicationVersion -- What version of the Cromwell application to deploy
```
choover-broad, line 3: Cromwell reference

charts/ontology/README.md
rushtong: Thank you! I have cleaned it up a little using the helm-docs tool, very helpful!

PR #100, opened 2020-08-06
DDO-466 jvm cromwell k8s

mflinn-broad:
This pr aims to enable jvm monitoring via prometheus for cromwell running in k8s. To do this we will run [prometheus-jmx exporter](https://github.com/prometheus/jmx_exporter) as a javaagent. It will expose a small http server which can be scraped for jvm metrics by a prometheus server.


charts/cromwell/templates/serviceMonitor.yaml
```diff
@@ -0,0 +1,18 @@
...
+  labels: 
+{{ include "cromwell.labels" . | indent 4 }}
+spec:
+  jobLabel: cromwell-jvm
+  selector:
+    matchLabels:
+      prometheus.io/scrape: "true"
+  namespaceSelector:
+    matchNames:
+    - terra-dev
```
gmalkov, line 14: Do we only want this for dev?

mflinn-broad: Oh I'll take this out. This was from before updated the prometheus config to support a more general pattern. Not needed anymore.

mflinn-broad: Just tested with the namespace selector removed and confirmed everything still works as intended

charts/cromwell/templates/metricsService.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: cromwell-metrics
```
choover-broad, line 4: With a Service and ServiceMonitor for all pods, is it possible to differentiate between metrics for the different deployments? Eg. cromwell1-reader vs cromwell1-runner vs cromwell2?

Having a single Service collect for all pods in multiple Deployments _feels_ a little surprising/unusual to me, but it very well might be a Standard Prometheus Pattern. I definitely don't think it'll break anything!

Would it make sense to use PodMonitors instead, so we don't have to create a new Service?

https://github.com/prometheus-operator/prometheus-operator/issues/3119#issuecomment-610445145

charts/cromwell/templates/_deployment.tpl
```diff
@@ -152,4 +167,11 @@ spec:
           subPath: sqlproxy-service-account.json
           name: sqlproxy-ctmpls
           readOnly: true
+      initContainers:
+      - name: download-prometheusjmx-jar
+        image: alpine:3.12.0
+        command: ["wget", "-O", "/cromwell-prometheusjmx-jar/prometheusjmx.jar", "https://repo1.maven.org/maven2/io/prometheus/jmx/jmx_prometheus_javaagent/0.13.0/jmx_prometheus_javaagent-0.13.0.jar"]
```
choover-broad, line 173: Would like it if this jar could be baked into, like, a standard monitoring init container image that is used by multiple charts. But it definitely doesn't need to be done right away/before merging

mflinn-broad: I agree, this would be the best approach long term. Will make a separate ticket to do this.

charts/cromwell/templates/metricsService.yaml
mflinn-broad: I did it this way because the service monitor is the standard pattern for prometheus. It can definitely still distinguish between all the different pods and cromwell service type. [Example](https://prometheus.dsde-dev.broadinstitute.org/graph?g0.range_input=1h&g0.expr=jvm_threads_current&g0.tab=0). 

We could use pod monitors if you feel strongly about it. That would require updating some configs on the prometheus side. How prometheus handles pod monitors and serviceMonitors are configured separately.

Pod monitor imo is more intended for when you want to correlate metrics across multiple systems. Something like appending a label with a trace id to metrics from a service and a database if both were running in kubernetes.

mflinn-broad: IMO the service monitor pattern is particularly well suited for Cromwell because it enables us to easily group metrics from a crowell deployment as a whole while retaining the ability to drill down to the individual cromwell components. With pod monitor we would lose that ability to easily correlate metrics across a whole cromwell deployment as a each pod would be considered a separate entity.

choover-broad: Nice! It looks like prometheus has included the pod name as a dimension no those metrics. Would it be possible to include the deployment type as a dimension as well?

Like, if I wanted to have a dashboard with just cromwell1 runner JVM metrics and exclude metrics for the other pods, could I?

mflinn-broad: yes we the dashboard I made supports this. Probably easier to just show you if you are interested.

mflinn-broad:
Converted to podMonitor pattern and confirmed it is functioning on dev prometheus instance

charts/cromwell/templates/jmxExporterCm.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: jmx-exporter-cm
```
choover-broad, line 4: Prefix this configmap name with `cromwell-` so it doesn't collide with any other Terra apps in same namespace?

choover-broad: One more thought -- does it make sense to include this in the regular Cromwell configmap so that changes to it trigger an automatic pod restart?

I think it would just mean changing the Java arg to refer to /etc/cromwell-cm/prometheusJmx.yaml instead

mflinn-broad: good idea. Will make this change

PR #101, opened 2020-08-07
[PF-61]: Enable pubsub in k8s by default

yonghaoy:


charts/crljanitor/values.yaml
```diff
@@ -39,6 +39,8 @@ serviceFirewallEnabled: false
 # serviceAllowedAddresses -- A map of addresses in the form `{ "nickname": ["x.x.x.x/y", "x.x.x.x/y"] }`
 serviceAllowedAddresses: {}
 
+# trackResourcePubsubEnabled -- Whether to enable track resource using pubsub.
```
wchamber, line 42: nit: Whether to enable using pubsub to receive new tracked resources.

PR #102, opened 2020-08-07
[PF-61]  Add app-SA and set it as the default credential for the app. Also rename the existing app-sa to cloudsql

yonghaoy:
Renamed the existing app-sa because we may have a dedicated SA for cloudsql

### DIFF
--- a/charts/crljanitor/templates/secrets.yaml
+++ b/charts/crljanitor/templates/secrets.yaml
@@ -56,6 +56,7 @@ spec:
       path: {{ .Values.vault.pathPrefix }}/crl_janitor/postgres/instance
       key: name
 ---
+# TODO: Create another service account in terraform for cloudsql proxy.
 apiVersion: secrets-manager.tuenti.io/v1alpha1
 kind: SecretDefinition
 metadata:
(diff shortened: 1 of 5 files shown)


charts/crljanitor/templates/secrets.yaml
```diff
@@ -59,11 +59,11 @@ spec:
...
-  name: secretdefinition-crl-janitor-app-sa
+  name: secretdefinition-crl-janitor-cloudsql-sa
   labels:
     {{ template "crljanitor.labels" . }}
 spec:
-  name: crl-janitor-sa
+  name: crl-janitor-cloudsql-sa
   keysMap:
     service-account.json:
       path: {{ .Values.vault.pathPrefix }}/crl_janitor/app-sa
```
wchamber, line 69: Is there a TODO here to create another service account in terraform and key in vault for this?

yonghaoy: Added a TODO. @gmalkov Hi Greg, I remembered you mentioned you are refactoring the SA to cloudsql-SA at some point? When that would be happen? Thanks

PR #104, opened 2020-08-10
PF-3 Set container and namespace env variables from k8s metadata for stackdriver stats exporting

wchamber:
To be used in https://github.com/DataBiosphere/terra-resource-janitor/pull/36

https://github.com/census-instrumentation/opencensus-java/blob/5be70440b53815eec1ab59513390aadbcec5cc9c/contrib/resource_util/src/main/java/io/opencensus/contrib/resource/util/ContainerResource.java#L79
https://github.com/census-instrumentation/opencensus-java/blob/5be70440b53815eec1ab59513390aadbcec5cc9c/contrib/resource_util/src/main/java/io/opencensus/contrib/resource/util/K8sResource.java#L112

### DIFF
--- a/charts/crljanitor/templates/deployment.yaml
+++ b/charts/crljanitor/templates/deployment.yaml
@@ -75,7 +75,7 @@ spec:
         - name: CONTAINER_NAME
           valueFrom:
             fieldRef:
-              fieldPath: metadata.name
+              fieldPath: metadata.labels['app']
         - name: NAMESPACE
           valueFrom:
             fieldRef:
--- a/charts/crljanitor/templates/deployment.yaml
+++ b/charts/crljanitor/templates/deployment.yaml
@@ -71,6 +71,15 @@ spec:
             secretKeyRef:
               name: crl-janitor-track-resource-pubsub
               key: subscription
+        # Environment variables for OpenCensus resource auto detection.
+        - name: CONTAINER_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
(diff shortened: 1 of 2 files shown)


charts/crljanitor/templates/deployment.yaml
```diff
@@ -71,6 +71,15 @@ spec:
...
               key: subscription
+        # Environment variables for OpenCensus resource auto detection.
+        - name: CONTAINER_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.labels['app']
+        - name: NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
```
mflinn-broad, line 82:
```suggestion
        - name: NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
        - name: VERSION
          value: {{ .Values.appVersion }}
```
It may be worth adding something like this so that an openCensus tag with the application version can be added to any metrics. I'm not sure if  Opencensus supports automatically extracting this value as with the above, but this is a common pattern when collecting metric data.

mflinn-broad: Discussed offline. It appears there is a system_labels.container_image field in stackdriver that is collected however it contains a full gcr path so may be problematic to work with. We are going to try the existing container image field first and may revisit this if that proves unmanageable.

wchamber:
> also update the version?

thanks, my version change got subsumed by another chart change. bumped again.

PR #108, opened 2020-08-26
[PF-9] Add configs in Janitor to support admin user authZ.

yonghaoy:
Those two configs are:

1. CONFIG_BASED_AUTHZ_ENABLED - boolean value to control whether this feature is enabled
2. ADMIN_USER_LIST - List of Janitor admin users

### DIFF
--- a/charts/crljanitor/values.yaml
+++ b/charts/crljanitor/values.yaml
@@ -44,9 +44,9 @@ serviceAllowedAddresses: {}
 
 # trackResourcePubsubEnabled -- Whether to enable using pubsub to receive new tracked resources.
 trackResourcePubsubEnabled: true
-# configBasedAuthZEnabled -- Whether to use static file with admin users email for authorization.
+# configBasedAuthzEnabled -- Whether to use static file with admin users email for authorization.
 # TODO(PF-81): Switch to use SAM instead of config file for user authZ.
-configBasedAuthZEnabled: true
+configBasedAuthzEnabled: true
 # DNS/TLS Values
 domain:
   # domain.hostname -- Hostname of this deployment
--- a/charts/crljanitor/values.yaml
+++ b/charts/crljanitor/values.yaml
@@ -29,6 +29,9 @@ vault:
   pathPrefix:
   # vault.configPathPrefix -- (string) Vault path prefix for configs. Required if vault.enabled.
   configPathPrefix:
+  # vault.adminUserFilePath -- (string) Vault path prefix for admin user list. Required if vault.enabled.
+  # Use the same users as admin for all env by default. Override if needed in helmfile repo.
+  adminUserFilePath: config/terra/crl-janitor/common/iam
 
 # serviceIP -- (string) External IP of the service. Required.
 serviceIP:
--- a/charts/crljanitor/values.yaml
+++ b/charts/crljanitor/values.yaml
@@ -44,6 +44,9 @@ serviceAllowedAddresses: {}
 
 # trackResourcePubsubEnabled -- Whether to enable using pubsub to receive new tracked resources.
 trackResourcePubsubEnabled: true
+# configBasedAuthZEnabled -- Whether to use static file with admin users email for authorization.
+# TODO(PF-81): Switch to use SAM instead of config file for user authZ.
+configBasedAuthZEnabled: true
 # DNS/TLS Values
 domain:
   # domain.hostname -- Hostname of this deployment
--- a/charts/crljanitor/Chart.yaml
+++ b/charts/crljanitor/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: crljanitor
-version: 0.1.10
+version: 0.1.11
 
 description: Chart for Cloud Resource Manager Janitor
 type: application
--- a/charts/crljanitor/templates/deployment.yaml
+++ b/charts/crljanitor/templates/deployment.yaml
@@ -74,7 +74,7 @@ spec:
         - name: GOOGLE_APPLICATION_CREDENTIALS
           value: /secrets/app/service-account.json
         - name: CONFIG_BASED_AUTHZ_ENABLED
-          value: "{{ .Values.configBasedAuthZEnabled }}"
+          value: "{{ .Values.configBasedAuthzEnabled }}"
         - name: ADMIN_USER_LIST
           valueFrom:
             secretKeyRef:
--- a/charts/crljanitor/templates/deployment.yaml
+++ b/charts/crljanitor/templates/deployment.yaml
@@ -73,6 +73,11 @@ spec:
               key: subscription
         - name: GOOGLE_APPLICATION_CREDENTIALS
           value: /secrets/app/service-account.json
+        - name: ADMIN_USER_LIST
+          valueFrom:
+            secretKeyRef:
+              name: crl-janitor-admin-user-list
+              key: admin-users
         # Environment variables for OpenCensus resource auto detection.
         - name: CONTAINER_NAME
           valueFrom:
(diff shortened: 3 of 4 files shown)


charts/crljanitor/values.yaml
```diff
@@ -29,6 +29,9 @@ vault:
   pathPrefix:
   # vault.configPathPrefix -- (string) Vault path prefix for configs. Required if vault.enabled.
   configPathPrefix:
+  # vault.adminUserFilePath -- (string) Vault path prefix for admin user list. Required if vault.enabled.
+  # Use the same users as admin for all env by default. Override if needed in helmfile repo.
+  adminUserFilePath: config/terra/crl-janitor/common/iam
```
gjuggler, line 34: I'm a little confused about the nature of this chart value. Is this path a Vault path, or a file path accessed from Java within the Docker container? The docs suggest the latter, but I think it's more the former.

Possible var names which might be clearer:
- adminUserSecretPath
- adminUserList
- adminUserListSecret

charts/crljanitor/values.yaml
```diff
@@ -41,6 +44,9 @@ serviceAllowedAddresses: {}
 
 # trackResourcePubsubEnabled -- Whether to enable using pubsub to receive new tracked resources.
 trackResourcePubsubEnabled: true
+# configBasedAuthZEnabled -- Whether to use static file with admin users email for authorization.
+# TODO(PF-81): Switch to use SAM instead of config file for user authZ.
+configBasedAuthZEnabled: true
```
gjuggler, line 49: super naming nit (feel free to punt if you disagree): the env variable treats "authz" as one token, but the capitalization here treats "z" as its own token. IMO it's more consistent to keep "authz" as one token, which would make this config name "configbasedAuthzEnabled" (also in the corresponding PR).

charts/crljanitor/values.yaml
yonghaoy: Yes, this is Vault path. And the path is , in that path, there would be bunch of key value paris inside. I plan to put all future potential IAM realated configs inside. So I name it as `config/terra/crl-janitor/common/iam`. It would have adminUser, SAM related, IC related things. 
Currently, inside `config/terra/crl-janitor/common/iam` path, there is a key adminUsers and the value is our config.

charts/crljanitor/values.yaml
yonghaoy: Done. Thanks!

yonghaoy:
> Straightforward enough!
> 
> One side question is whether we need to treat the admin user list as a secret at all, or if we're happy enough having it specified directly within our Helm chart(s) somewhere. The fact that we're placing it within the "/config" Vault namespace suggests that maybe we're OK with the source of truth being somewhere in Git? (Not saying it's worth changing at the moment, but it might be helpful to know whether that would be an ideal setup in the long run, and which repo would be the right source of truth for that if it were.

Thanks @gjuggler . In the configs doc: https://docs.google.com/document/d/1esnGc16v6hghMmoI_9uTyyDFkJSG84K6G5R9zn8cYho/edit#heading=h.5jnyigd46ehx, I had the same feeling and listed in the problem2 'The boundary of configs and secrets are not clear'. I think the ideal solution is to use k8s secret manager to secret, and use ConfigMap(https://kubernetes.io/docs/concepts/configuration/configmap/) to manage k8s configs. But currently we put everything in k8s secret. Because the terraform-k8s piping have to use 'secret manager - k8s secret path'. So we put pubsub topic, project id in secret manager even through I feel they should all be configs. 
We can check with @gmalkov on next devOps sync

gjuggler:
> > Straightforward enough!
> > One side question is whether we need to treat the admin user list as a secret at all, or if we're happy enough having it specified directly within our Helm chart(s) somewhere. The fact that we're placing it within the "/config" Vault namespace suggests that maybe we're OK with the source of truth being somewhere in Git? (Not saying it's worth changing at the moment, but it might be helpful to know whether that would be an ideal setup in the long run, and which repo would be the right source of truth for that if it were.
> 
> Thanks @gjuggler . In the configs doc: https://docs.google.com/document/d/1esnGc16v6hghMmoI_9uTyyDFkJSG84K6G5R9zn8cYho/edit#heading=h.5jnyigd46ehx, I had the same feeling and listed in the problem2 'The boundary of configs and secrets are not clear'. I think the ideal solution is to use k8s secret manager to secret, and use ConfigMap(https://kubernetes.io/docs/concepts/configuration/configmap/) to manage k8s configs. But currently we put everything in k8s secret. Because the terraform-k8s piping have to use 'secret manager - k8s secret path'. So we put pubsub topic, project id in secret manager even through I feel they should all be configs.
> We can check with @gmalkov on next devOps sync

Thanks. The root of my question was a bit deeper: do we need to use Vault for the admin user list at all? e.g. we could treat the adminUserList as a Helm value and specified the set of allowed Broad admins in the terra-helmfile repo, e.g. https://github.com/broadinstitute/terra-helmfile/blob/master/terra/values/crljanitor.yaml.

I can imagine a couple arguments for actively wanting to use Vault – for example, we don't want Broad email addresses to be tracked in GitHub (even if it's in a private repo), or maybe it's such a common pattern used by the Broad that it doesn't make sense to try something different.

Pragmatically speaking, it's almost certainly not worth a change. But I just want to better understand how we reason about these things.

yonghaoy:
> > > Straightforward enough!
> > > One side question is whether we need to treat the admin user list as a secret at all, or if we're happy enough having it specified directly within our Helm chart(s) somewhere. The fact that we're placing it within the "/config" Vault namespace suggests that maybe we're OK with the source of truth being somewhere in Git? (Not saying it's worth changing at the moment, but it might be helpful to know whether that would be an ideal setup in the long run, and which repo would be the right source of truth for that if it were.
> > 
> > 
> > Thanks @gjuggler . In the configs doc: https://docs.google.com/document/d/1esnGc16v6hghMmoI_9uTyyDFkJSG84K6G5R9zn8cYho/edit#heading=h.5jnyigd46ehx, I had the same feeling and listed in the problem2 'The boundary of configs and secrets are not clear'. I think the ideal solution is to use k8s secret manager to secret, and use ConfigMap(https://kubernetes.io/docs/concepts/configuration/configmap/) to manage k8s configs. But currently we put everything in k8s secret. Because the terraform-k8s piping have to use 'secret manager - k8s secret path'. So we put pubsub topic, project id in secret manager even through I feel they should all be configs.
> > We can check with @gmalkov on next devOps sync
> 
> Thanks. The root of my question was a bit deeper: do we need to use Vault for the admin user list at all? e.g. we could treat the adminUserList as a Helm value and specified the set of allowed Broad admins in the terra-helmfile repo, e.g. https://github.com/broadinstitute/terra-helmfile/blob/master/terra/values/crljanitor.yaml.
> 
> I can imagine a couple arguments for actively wanting to use Vault – for example, we don't want Broad email addresses to be tracked in GitHub (even if it's in a private repo), or maybe it's such a common pattern used by the Broad that it doesn't make sense to try something different.
> 
> Pragmatically speaking, it's almost certainly not worth a change. But I just want to better understand how we reason about these things.

I checked with @gmalkov about what is the current best practice in our use case. And the conclusion is to use Vault and give a github team(janitor admin) modify permission. The the value is 
(1) Configs can only be viewed for the people with Vault read access
(2) We can prevent any random person make PR change to add random people. Currently more people than we expect can merge PR in helm repo(I think all DSP Engineers).
(3) Not sure making a PR change then deploy the application is the best way to upgrade permission.

I agree in the long term we need a better piping workflow for configs. Currently most hard coded values in helm values are Vault path, and all actual values are in Vault. I would assume we will get rid of Vault, (Similar to Vault but some other central places with all hard coded values and people can have different permissions on different path). And this piping workflow is able to update k8s ConfigMap accordingly. I believe that is on the roadmap of DevOps.

PR #109, opened 2020-08-28
DDO-662 Add support for jvm monitoring via prometheus

mflinn-broad:
This pr enables jvm profiling for k8s deployents of the ontology service. This is accomplished by running the [prometheus-jmx-exporter](https://github.com/prometheus/jmx_exporter) as a javaagent. This will expose detailed jvm profiling metrics which will then be sent to stackdriver by the prometheus deployment in the terra clusters.

### DIFF
--- a/charts/ontology/templates/deployment.yaml
+++ b/charts/ontology/templates/deployment.yaml
@@ -54,6 +54,9 @@ spec:
               mountPath: /etc/service-account.json
               subPath: service-account.json
               readOnly: true
+            - name: ontology-prometheus-jmx-jar
+              mountPath: /etc/prometheusjmx/prometheusjmx.jar
+              subPath: prometheusjmx.jar
           readinessProbe:
             httpGet:
               path: /status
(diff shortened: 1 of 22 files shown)


mflinn-broad:
I changed the way java args are passed in order to allow the javaagent arg to be parsed correctly. This is also the same pattern we use in the [cromwell chart](https://github.com/broadinstitute/terra-helm/blob/846db0db5d0c73c4c1f220f3e5303d292f8f7c45/charts/cromwell/templates/_deployment.tpl#L66)

charts/ontology/templates/deployment.yaml
```diff
@@ -43,17 +46,27 @@ spec:
                 secretKeyRef:
                   name: {{ .Chart.Name }}-secrets
                   key: dsn
-          command: ["java"]
```
choover-broad, line 46: IMO it's actually better to invoke `java` directly here instead of using Bash :) This way we're not expecting a particular shell to exist on the container image. We should probably update the Cromwell chart to do the same (if it works)

mflinn-broad: Fair point. I can try to get it to work switching back to invoking java directly. May need to play around with how we pass arguments in that case.  I was getting `unsupported argurment` errors when trying to pass the `-javaagent` arg when using the java command directly.

mflinn-broad: I'll play around with it some more to see if I can get it working with the java command pattern.

mflinn-broad: Might be able to pass the `-javaagent` as an env var or something

rushtong: Agree with Chelsea here.

PR #111, opened 2020-08-31
IA-2164 add configs for resource-validator

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2164


charts/leonardo/templates/_configMap.tpl
```diff
@@ -0,0 +1,27 @@
+{{- /* Generate the data component of a Cromwell config map */ -}}
```
Qi77Qi, line 1: these are not used yet...But probably will be needed for Leonardo, hence leaving it there for now

charts/leonardo/templates/cronjob.yaml
```diff
@@ -0,0 +1,91 @@
...
+            - mountPath: /tmp/pod
+              name: tmp-pod
+              readOnly: true
+          volumes:
+          - name: leonardo-sa-json-file
+            secret:
+              secretName: leonardo-sa-json
+          - name: tmp-pod
+            emptyDir: {}
+          - name: logback-config
```
Qi77Qi, line 86: not needed for cron job..but can be useful for later migrating Leonardo, hence leaving it here for now

charts/leonardo/templates/cronjob.yaml
```diff
@@ -0,0 +1,91 @@
...
+                  name: leonardo-db-creds
+                  key: instance_region
+            - name: SQL_INSTANCE_NAME
+              valueFrom:
+                secretKeyRef:
+                  name: leonardo-db-creds
+                  key: instance_name
+            command: ["/bin/sh", "-c"]
+            args:
+            - |
```
choover-broad, line 67: Add comment indicating why this workaround is here & that it should be removed once sidecar lifecycle support is available?

charts/leonardo/templates/roleBinding.yaml
```diff
@@ -1,12 +1,12 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
-  name: leonardo-sa-binding
+  name: leonardo-ksa-binding
```
choover-broad, line 4: Should this be `leonardo-sa` / `leonardo-sa-binding` to stick with conventions in other charts? Especially since the custom psp is no longer needed

charts/leonardo/values.yaml
```diff
@@ -39,3 +39,22 @@ deployments:
     expose: true
     # deployments.standalone.serviceName -- Name of the default standalone Leonardo service
     serviceName: leonardo
+
+cronjob:
+  # deployments.standalone.name -- Name to use for the default standalone Leonardo deployment
+  name: resource-validator-cronjob
+  # deployments.standalone.replicas -- Number of replicas in the default standalone Leonardo deployment
+  replicas: 1
```
choover-broad, line 47: under `cronjob`, remove the `name`, `replicas`, `expose`, `serviceName`, and `enabled` values, as they aren't used by the cronjob template

charts/leonardo/values.yaml
```diff
@@ -39,3 +39,22 @@ deployments:
...
+  expose: false
+  # deployments.standalone.serviceName -- Name of the default standalone Leonardo service
+  serviceName: resourceValidator
+  enabled: true
+  imageRepository: us.gcr.io/broad-dsp-gcr-public/resource-validator
+  imageTag: 0.0.1-rc2
+
+vault:
+  # vault.enabled -- When enabled, syncs required secrets from Vault
+  enabled: true
```
choover-broad, line 58: remove `vault.enabled`, chart ignores it + cronjob won't work if it is disabled

charts/leonardo/templates/cronjob.yaml
```diff
@@ -0,0 +1,91 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: "resource-validator-cron"
```
choover-broad, line 4: Can this be `leonardo-resource-validator-cron`?

In general, every K8s resource that is deployed via the leonardo chart should be prefixed with `leonardo`. All the app charts for an environment are deployed side-by-side in the same namespace (like `terra-dev`), and having a common prefix for a chart's resources reduces likelihood of conflicts

Sorry, we definitely need to document best practices like this somewhere :/

charts/leonardo/templates/cronjob.yaml
```diff
@@ -0,0 +1,91 @@
...
+              value: /secrets/service-account.json
+            volumeMounts:
+            - name: leonardo-sa-json-file
+              mountPath: /secrets
+              readOnly: true
+            - mountPath: /tmp/pod
+              name: tmp-pod
+            - mountPath: /opt/docker/resource-validator-config
+              name: logback-config
+          - name: cloudsql-proxy
```
choover-broad, line 47: Container names are used to filter log messages in Stackdriver, so could this be `leonardo-resource-validator-cloudsql-proxy` or similar?

(Again, sorry, we really need to document best practices)

charts/leonardo/templates/cronjob.yaml
```diff
@@ -0,0 +1,91 @@
...
+        spec:
+          restartPolicy: Never
+          serviceAccountName: leonardo-ksa
+          containers:
+          - image: "{{ .Values.cronjob.imageRepository }}:{{ .Values.cronjob.imageTag }}"
+            name: "{{ .Values.cronjob.name }}"
+            command: ["/bin/bash", "-c"]
+            args:
+              - |
+                trap "touch /tmp/pod/main-terminated" EXIT
```
choover-broad, line 22: Add comment explaining why this is here?

charts/leonardo/templates/cronjob.yaml
choover-broad: Actually... Will the resource-validator and the leonardo app share any config files in common? 🤔

choover-broad (changes requested):
@Qi77Qi -- nice work figuring out how to get CronJobs working with a sidecar, you're breaking new ground for us here!

I left a few style & naming comments in-line, most are related to best practices that we unfortunately have not documented yet :(

I think that the ConfigMap stuff should just be removed (unless you think that the CronJob and Leonardo app will have any non-secret configuration files in common). It will most likely have to be rewritten anyway as the Leonardo app is ported over.

charts/leonardo/templates/roleBinding.yaml
Qi77Qi: In leonardo team, we refer to kubernetes sa as ksa, and normal google service account as sa...I think it's confusing when we use sa here, because most people know it as google service account instead of ksa. And in the chart, we do refer to leonardo's app service account as Leonardo sa.

choover-broad: yeah, I hear that... But I think it's valuable to have things follow a common standard. It'll be a bit awkward if we have `sam-sa`, `rawls-sa`, `cromwell-sa`, and then, randomly, `leonardo-ksa`... As a DevOps engineer inspecting a Terra Kubernetes namespace, I'd be wondering if `leonardo-ksa` is intended for a different purpose than these other application SAs

Qi77Qi: reverted

choover-broad (approved):

PR #115, opened 2020-09-10
[AS-484] Add tcell, LDAP support, ID whitelisting to WM proxy

zloery:
See the [wiki page](https://broadinstitute.atlassian.net/wiki/spaces/GAWB/pages/1589608449/Apache+Proxy) for more background on the proxy.

This adds the configuration and settings to use TCell in Workspace Manager's proxy, to perform LDAP lookups for authorization, and to check OIDC audience whitelists.

Based heavily on existing ctmpl proxy configurations: see [Rawls configuration](https://github.com/broadinstitute/firecloud-develop/blob/dev/base-configs/rawls/site.conf.ctmpl) with [values file](https://github.com/broadinstitute/firecloud-develop/blob/dev/run-context/live/configs/rawls/proxy-compose.yaml.ctmpl).

Corresponding changes for values files are in [this helmfile PR](https://github.com/broadinstitute/terra-helmfile/pull/182)

### DIFF
--- a/charts/workspacemanager/templates/_helpers.tpl
+++ b/charts/workspacemanager/templates/_helpers.tpl
@@ -44,3 +44,52 @@ LDAP base domain template
 {{- define "workspacemanager.ldapbasedomain" -}}
   dc=dsde-{{ .Values.global.terraEnv }},dc=broadinstitute,dc=org
 {{- end }}
+
+{{/*
+AoU Preprod Conditional
+*/}}
+{{- define "workspacemanager.aoupreprodconditional" -}}
+  ((.email|endswith("@preprod.researchallofus.org")) and (.audience | split(".")[0] as $idkey | "${AOU_PREPROD_WHITELIST}" | split(",") | any(contains($idkey))))
+{{- end }}
+{{/*
+AoU Perf Conditional
+*/}}
+{{- define "workspacemanager.aouperfconditional" -}}
+  ((.email|endswith("@perf.fake-research-aou.org")) and (.audience | split(".")[0] as $idkey | "${AOU_PERF_WHITELIST}" | split(",") | any(contains($idkey))))
+{{- end }}
+{{/*
+AoU Prod Conditional
+*/}}
+{{- define "workspacemanager.aouprodconditional" -}}
+  ((.email|endswith("@researchallofus.org")) and (.audience | split(".")[0] as $idkey | "${AOU_PROD_WHITELIST}" | split(",") | any(contains($idkey))))
+{{- end }}
+{{/*
+AoU Dev Conditional
+*/}}
+{{- define "workspacemanager.aoudevconditional" -}}
+  ((.email|endswith("@fake-research-aou.org")) and (.audience | split(".")[0] as $idkey | "${AOU_DEV_WHITELIST}" | split(",") | any(contains($idkey))))
+{{- end }}
+{{/*
+AoU Stable Conditional
+*/}}
+{{- define "workspacemanager.aoustableconditional" -}}
+  ((.email|endswith("@stable.fake-research-aou.org")) and (.audience | split(".")[0] as $idkey | "${AOU_STABLE_WHITELIST}" | split(",") | any(contains($idkey))))
+{{- end }}
+{{/*
+AoU Staging Conditional
+*/}}
+{{- define "workspacemanager.aoustagingconditional" -}}
+  ((.email|endswith("@staging.fake-research-aou.org")) and (.audience | split(".")[0] as $idkey | "${AOU_STAGING_WHITELIST}" | split(",") | any(contains($idkey))))
+{{- end }}
+{{/*
+Terra ID Whitelist
+*/}}
+{{- define "workspacemanager.terraidwhitelist" -}}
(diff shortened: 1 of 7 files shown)


zloery:
My thoughts on the whitelisting: this is harder to do in helm because the template command for evaluating environment variables is disabled, meaning I can't work with the whitelist values in the template.

One workaround that might work is storing the whitelist IDs as a regex pattern instead of a JSON string - something like "\bID_1\b|\bID_2\b|...." - and using the proxy's regex matching. That feels a little hacky and I'm not thrilled about storing the ID whitelist that way, so I want to get other people's thoughts on this problem before committing to that solution.

zloery:
Another (probably easier) option would be to store the whitelist IDs in a different format than JSON, like a comma separated list. Because `claims_expr` lets us run `jq`, we could do basic string manipulation there.

charts/workspacemanager/templates/_helpers.tpl
```diff
@@ -37,3 +37,59 @@ Service firewall
   {{- end -}}
   {{- end -}}
 {{- end }}
+
+{{/*
+LDAP base domain template
+*/}}
+{{- define "workspacemanager.ldapbasedomain" -}}
+  dc=dsde-{{ .Values.global.terraEnv }},dc=broadinstitute,dc=org
```
gmalkov, line 45: We are trying to steer away from using global.terraEnv in string interpolations and instead using explicit values. Using `global.terraEnv` assumes/forces certain naming conventions and makes these charts less portable. For example, in this case I would recommend replacing this template with a `.Values.proxy.ldap.baseDomain` value. This will cause our values overrides to become more verbose, but we think it's worth it in the end.

charts/workspacemanager/templates/deployment.yaml
```diff
@@ -125,6 +125,75 @@ spec:
...
+            secretKeyRef:
+              name: proxy-oauth-whitelist
+              key: terra
+        - name: REMOTE_USER_CLAIM
+          value: sub
+        {{- if .Values.proxy.ldap.enabled }}
+        - name: LDAP_BASE_DOMAIN
+          value: {{ template "workspacemanager.ldapbasedomain" . }}
+        - name: AUTH_LDAP_URL2
+          value: 'AuthLDAPURL "ldaps://opendj.dsde-{{ .Values.global.terraEnv }}.broadinstitute.org/ou=people,{{ template "workspacemanager.ldapbasedomain" . }}?googleSubjectId?sub?(objectClass=*)"'
```
gmalkov, line 170: Also recommend avoiding use of `.Values.global.terraEnv` here
```suggestion
          value: 'AuthLDAPURL "ldaps://{{ .Values.proxy.ldap.url }}/ou=people,{{ template "workspacemanager.ldapbasedomain" . }}?googleSubjectId?sub?(objectClass=*)"'
```

charts/workspacemanager/templates/tcell_configmap.yaml
```diff
@@ -0,0 +1,24 @@
...
+    {
+      "version": 1,
+      "applications": [
+        {
+          "app_id": "$(TCELL_APP_ID)",
+          "api_key": "$(TCELL_API_KEY)",
+          "tcell_api_url": "https://us.agent.tcell.insight.rapid7.com/api/v1",
+          "tcell_input_url": "https://us.input.tcell.insight.rapid7.com/api/v1",
+          "js_agent_api_base_url": "https://us.agent.tcell.insight.rapid7.com/api/v1",
+          "host_identifier":"workspace-manager-{{ .Values.global.terraEnv }}",
```
gmalkov, line 19: Same deal, try and avoid using `.Values.global.terraEnv`

gmalkov (commented):
Looks great! I am a little fuzzy on the whitelisting stuff, so maybe we can schedule a quick call and you can walk me through it?

zloery:
> Another (probably easier) option would be to store the whitelist IDs in a different format than JSON, like a comma separated list. Because `claims_expr` lets us run `jq`, we could do basic string manipulation there.

I ended up going with this option. I added a field to each whitelist secret with the IDs stored as single comma separated string, and the current code uses those strings for the whitelist values.

gmalkov:
> I ended up going with this option. I added a field to each whitelist secret with the IDs stored as single comma separated string, and the current code uses those strings for the whitelist values.
Gotcha, OK, makes sense!

gmalkov (approved):

gmalkov:
Remember to bump the chart version before merging

zloery:
Rendered the site.conf to make things a little easier to compare (Vault secrets aren't rendered, and actual whitelist values are modified):

- [Workspace Manager's config](https://docs.google.com/document/d/16T22AjHMeI4_AHL9x4fzmkbzpdkFGHbTT5650gAx0Es/edit?usp=sharing)
- [Current dev Rawls config](https://docs.google.com/document/d/1WUdeAIUap4RwwOHsRTVi8p1vPEi-gv_JeDGLKiW9KkE/edit?usp=sharing)
- [Diff as rendered by google docs](https://docs.google.com/document/d/1du4D0Fe-ZnygA8bF99dD8CI4FCXPD9AnOnnnsoR9ejo/edit?usp=sharing)

dvoet (approved):
1 minor diff noted in one of the docs but LGTM

PR #123, opened 2020-09-17
Add Revision History Limit

rushtong:
Add a revision history limit to deployments. Not sure if this is the right place to add it, can't find much in the argo docs on it.


choover-broad (approved):
LGTM!

mflinn-broad (approved):

PR #137, opened 2020-09-25
[PF-110] initial rbs helm chart

yonghaoy:
Tried using skaffold, it works:
https://rbs.yyu.integ.envs.broadinstitute.org/status

### DIFF
--- a/charts/rbs/templates/deployment.yaml
+++ b/charts/rbs/templates/deployment.yaml
@@ -1,0 +1,138 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: terra-rbs-deployment
+  labels:
+    {{ template "rbs.labels" . }}
+spec:
+  replicas: {{ .Values.replicas }}
+  selector:
+    matchLabels:
+      app: terra-rbs
+  template:
+    metadata:
+      name: terra-rbs-service
+      labels:
+        app: terra-rbs
+    spec:
+      serviceAccountName: terra-rbs-service-sa
+      containers:
+      - name: terra-rbs
+        image: "{{- if .Values.image -}}
+            {{ .Values.image }}
+          {{- else -}}
+            {{ .Values.imageConfig.repository }}:{{ default .Values.appVersion .Values.imageConfig.tag }}
+          {{- end }}"
+        imagePullPolicy: {{ .Values.imageConfig.imagePullPolicy }}
+        {{- if not .Values.proxy.enabled }}
+        ports:
+        - containerPort: 8080
+        {{- end}}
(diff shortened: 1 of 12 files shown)


charts/rbs/templates/deployment.yaml
```diff
@@ -0,0 +1,138 @@
...
+spec:
+  replicas: {{ .Values.replicas }}
+  selector:
+    matchLabels:
+      app: terra-rbs
+  template:
+    metadata:
+      name: terra-rbs-service
+      labels:
+        app: terra-rbs
```
mflinn-broad, line 16: Do we want to include the other ```{{ template "rbs.labels" .  }}``` here in addition so that pods will have those labels too?

charts/rbs/templates/proxy-apache-httpd-configmap.yaml
```diff
@@ -0,0 +1,83 @@
...
+          ${AUTH_TYPE2}
+          ${AUTH_REQUIRE2}
+
+          ProxyPass http://localhost:8080/api
+          ProxyPassReverse http://localhost:8080/api
+
+          AddOutputFilterByType DEFLATE application/json text/plain text/html application/javascript application/x-javascript
+        </Location>
+    </VirtualHost>
+{{- end }}
```
mflinn-broad, line 83: We need to use the updated [apache config](https://github.com/broadinstitute/terra-helm/blob/master/charts/workspacemanager/templates/proxy-apache-httpd-configmap.yaml) that was added for workspace manager recently in this case too. There were some security issues that came up with the infosec team when reviewing this config. @gmalkov might have a little more context on this.

mflinn-broad (changes requested):
Looks fine apart from the issue with the apache config.

charts/rbs/templates/proxy-apache-httpd-configmap.yaml
yonghaoy: Thanks Mike! I will check with Greg. But for RBS, there is no user facing APIs, so I don't think we will need Ldap stuff, AoU precondition. Not sure the Tcell thing though

yonghaoy: Hi Mike, checked with Greg offline. RBS probably need the same stuff as WM. We will live the RBS as it, and switch later if we need.

charts/rbs/templates/deployment.yaml
yonghaoy: Done

mflinn-broad (approved):
lgtm

PR #146, opened 2020-10-01
DDO-758 Add liveness check to Cromwell chart

choover-broad:
Update Cromwell liveness/readiness checks so that:
* pod is taken out of service after 60s of failed polling
* pod is restarted after 300s of failed polling

Also use @gmalkov's awesome new workflow to auto-bump the Cromwell chart version in terra-helmfile


mflinn-broad (approved):
👍

PR #147, opened 2020-10-01
DDO-761 Ontology Liveness Probe

mflinn-broad:
In Addition to the existing readiness probe which will prevent an ontology pod from
receiving traffic if the status endpoint is unresponsive, this pr adds a liveness probe.
The liveness probe will automatically restart the ontology container after 5 minutes of status check
failues.
﻿- Add liveness probe
- bump chart version


choover-broad (approved):
Nice!

PR #148, opened 2020-10-01
Check if master tag or not based on tag regex

gmalkov:


choover-broad (approved):
Took a while to parse that regex, heh, but looks good! 👍

PR #151, opened 2020-10-06
[PF-112] Update environment var name to use spring auto environment variable

yonghaoy:
Related comment threads:
https://docs.google.com/document/d/1VfgL8D6PpJg7irbL_gKnmyZ5_GH5OYGfK8GPN45odUs/edit?disco=AAAAG95t27k
Spring variable name reference: https://github.com/DataBiosphere/terra-rbs/blob/master/src/main/resources/application.yml

### DIFF
--- a/charts/rbs/templates/deployment.yaml
+++ b/charts/rbs/templates/deployment.yaml
@@ -29,32 +29,32 @@ spec:
         - containerPort: 8080
         {{- end}}
         env:
-        - name: DATABASE_USER
+        - name: RBS_DB_USERNAME
           valueFrom:
             secretKeyRef:
               name: rbs-postgres-db-creds
               key: username
-        - name: DATABASE_USER_PASSWORD
+        - name: RBS_DB_PASSWORD
           valueFrom:
             secretKeyRef:
               name: rbs-postgres-db-creds
               key: password
-        - name: DATABASE_NAME
+        - name: RBS_DATABASE_NAME
           valueFrom:
             secretKeyRef:
               name: rbs-postgres-db-creds
               key: db
-        - name: STAIRWAY_DATABASE_USER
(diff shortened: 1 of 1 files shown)


charts/rbs/templates/deployment.yaml
```diff
@@ -29,32 +29,32 @@ spec:
...
               name: rbs-stairway-db-creds
               key: username
-        - name: STAIRWAY_DATABASE_USER_PASSWORD
+        - name: RBS_STAIRWAY_DB_PASSWORD
           valueFrom:
             secretKeyRef:
               name: rbs-stairway-db-creds
               key: password
-        - name: STAIRWAY_DATABASE_NAME
+        - name: RBS_STAIRWAY_DATABASE_NAME
```
gjuggler, line 57: side comment / not for this PR: I think we can eventually migrate this one to look more like the other variables, e.g. map directly to a Spring application property rather than be templated into the Spring config via environment variable templating. The trick would be to change our Spring application config in something like this way:

```
// Current app.yaml
rbs.stairway.db:
      password:
      uri: jdbc:postgresql://127.0.0.1:5432/${RBS_STAIRWAY_DATABASE_NAME}
      username:

// Future app.yaml
rbs.stairway.db:
      name: 
      password: 
      uri: jdbc:postgresql://127.0.0.1:5432/${rbs.stairway.db.name}
      username: 

// Future helm chart
- name: RBS_STAIRWAY_DB_NAME
          valueFrom:
            secretKeyRef:
              name: rbs-stairway-db-creds
              key: db

```


I think I still have something like this in a Janitor dev branch... I just need to get it prepped for review and to figure out how to make a coordinated change across the Helm file and the application version. I'll ping you once I'm ready to bring that one into review.

gjuggler (approved):

PR #154, opened 2020-10-13
DUOS-298 - Consent helm chart

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-298
See also https://github.com/broadinstitute/terra-helmfile/pull/343

Helm chart for consent


charts/consent/templates/deployment.yaml
```diff
@@ -0,0 +1,155 @@
...
+              subPath: site.conf
+              readOnly: true
+            - name: proxy-secrets
+              mountPath: "/etc/ssl"
+              readOnly: true
+        - name: {{ .Chart.Name }}-sqlproxy
+          image: broadinstitute/cloudsqlproxy:1.11_20191120
+          envFrom:
+            - configMapRef:
+                name: configs
```
mflinn-broad, line 116:
```suggestion
                name: consent-configmap
```
I think you are trying to reference the volume created from the config map on line 131.
This parameter expects the name of k8s config map rather than the volume I think.

charts/consent/templates/_consent.yaml.tpl
```diff
@@ -0,0 +1,63 @@
...
+  initialSize: 20
+  minSize: 20
+  maxSize: 128
+  validationQuery: SELECT 1
+elasticSearch:
+  servers:
+    - {{ .Values.elasticSearchServer1 }}
+    - {{ .Values.elasticSearchServer2 }}
+    - {{ .Values.elasticSearchServer3 }}
+  indexName: ontology
```
mflinn-broad, line 36: is this left over from ontology or still needed here?

charts/consent/templates/deployment.yaml
```diff
@@ -0,0 +1,177 @@
...
+  revisionHistoryLimit: 0
+  selector:
+    matchLabels:
+      app: {{ .Chart.Name }}
+  template:
+    metadata:
+      name: {{ .Chart.Name }}
+      labels:
+{{ include "consent.labels" . | indent 8 }}
+        app: {{ .Chart.Name }}
```
mflinn-broad, line 18: can we just add `app: {{ .Chart.Name }}` to the consent labels template. We should make this change for ontology too. I think it was  like this for ontology because we added the template labels later

Could also just `app.kubernetes.io/component: {{ .Chart.Name }}` in `matchLabels` on the selector

charts/consent/templates/deployment.yaml
```diff
@@ -0,0 +1,177 @@
...
+          readinessProbe:
+            httpGet:
+              path: /status
+              port: 8080
+            initialDelaySeconds: 60
+            periodSeconds: 10
+            timeoutSeconds: 1
+            failureThreshold: 10
+            successThreshold: 1
+          securityContext:
```
mflinn-broad, line 112:
```suggestion
          livenessProbe:
            httpGet:
              path: /status
              port: 8080
            timeoutSeconds: 5
            initialDelaySeconds: 60
            periodSeconds: 10
            failureThreshold: 30 # Restart container after 5 minutes
          securityContext:
```
Can we add the liveness probe that will restart the pod after 5 minutes of failed status checks also?

charts/consent/templates/podMonitor.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: PodMonitor
+metadata:
+  name: consent-jvm-monitor
+  labels:
+{{ include "consent.labels" . | indent 4 }}
+spec:
+  jobLabel: consent-jvm
+  selector:
+    matchLabels:
+      app.kubernetes.io/component: consent
```
mflinn-broad, line 11:
```suggestion
      app.kubernetes.io/component: {{ .Chart.name }}
```
Just in case 😄

mflinn-broad (changes requested):
A couple minor changes, looks great to me though, excited about this!

charts/consent/templates/_consent.yaml.tpl
rushtong: Currently, Consent also talks to the ES servers. Ontology just does read-only stuff, Consent does index file updating. I want to extract that feature from Consent in the future, but for now, it's  necessary.

mflinn-broad (approved):
👍

PR #162, opened 2020-10-19
fix chart name reference

rushtong:
Update the `Chart.Name` reference. For some reason, even though the key is named `name` with a lower case `n`, [it does not validate correctly](https://github.com/broadinstitute/terra-helmfile/pull/343/checks?check_run_id=1275364642):
```
  Error: template: consent/templates/podMonitor.yaml:11:44: executing "consent/templates/podMonitor.yaml" at <.Chart.name>: can't evaluate field name in type interface {}
```

### DIFF
--- a/charts/consent/Chart.yaml
+++ b/charts/consent/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: consent
-version: 0.2.0
+version: 0.3.0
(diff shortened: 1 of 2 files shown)


mflinn-broad (approved):
I always forget about what needs to be capitalized in the helm templates too.

PR #164, opened 2020-10-20
PF-152 Use the WM app SA as the default GOOGLE_APPLICATION_CREDENTIALS.

wchamber:
Change the helm for workspace manager to set the new app SA as the default Google SA with environment variable GOOGLE_APPLICATION_CREDENTIALS. This will match other MC Terra services.
This is a backwards compatible change with existing WM deployments: the app SA has a superset of the permissions in the cloud trace SA, including those needed for logging & monitoring in GKE. We also don't modify the cloudtrace SA volume or variables so that existing images can still use it.
Cleanup for the cloudtrace SA is tracked in PF-197.

### DIFF
--- a/charts/workspacemanager/Chart.yaml
+++ b/charts/workspacemanager/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: workspacemanager
-version: 0.7.0
+version: 0.8.0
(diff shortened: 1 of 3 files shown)


wchamber:
I tested with a skaffold run using this branch.

zloery (approved):

PR #165, opened 2020-10-20
[IA-2033] add application.conf all cron jobs

Qi77Qi:
Tested locally in my own namespace

1. Adds `application.conf` for each job for environment specific variables
2. Bump resource validator and zombie monitor to latest image
3. Add `leonardo-pubsub.google-project` to application.conf


charts/leonardo/values.yaml
```diff
@@ -43,7 +43,8 @@ cronjob:
   # deployments.standalone.name -- Name to use for the default standalone Leonardo deployment
   name: leonardo-resource-validator-cronjob
   imageRepository: us.gcr.io/broad-dsp-gcr-public/resource-validator
-  imageTag: b994ab5
+  imageTag: latest
```
kyuksel, line 46: Should we peg to a specific hash instead?

Qi77Qi: this needs to be updated once your PR is merged...I'm using `latest` temporarily because this set up won't work without https://github.com/broadinstitute/leonardo-cron-jobs/pull/17

Qi77Qi: oh actually, broadinstitute/leonardo-cron-jobs#17 will generate new images as well..I'll update here once those images are published

kyuksel (approved):

PR #169, opened 2020-10-21
No ticket Fix to /etc/ssl overwrite in consent proxy volume mounts

mflinn-broad:
The oidc proxy container has a bunch of files built into the image under `/etc/ssl` the pattern being used previously to mount tls credentials to the proxy container was overwriting the entire contents of `/etc/ssl` This pr adjust the mounting pattern for tls credentials so that they are inserted into the correct paths under `/etc/ssl/` rather than overwriting the entire directory.


rushtong (approved):
Looks good!

PR #176, opened 2020-10-23
[AS-524] Fix audience checking assumptions to use startswith

zloery:
Currently, the WM proxy audience check assumes that the first portion of an audience (split by "." characters) has a format like "12345678901". However, some valid audiences have a first portion format like "998877665544-abcdefghijkl54321blahblahblah". This check is rejecting valid calls from Rawls.

Current services check that audiences **start with** the whitelisted ids. This change makes WM do the same.

(I'd like to test this more before submitting, I'll bring up the new version in my integration env and we can test against that)

### DIFF
--- a/charts/workspacemanager/templates/_helpers.tpl
+++ b/charts/workspacemanager/templates/_helpers.tpl
@@ -42,43 +42,43 @@ Service firewall
 AoU Preprod Conditional
 */}}
 {{- define "workspacemanager.aoupreprodconditional" -}}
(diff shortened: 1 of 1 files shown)


davidangb (approved):
this looks correct from reading the code, just needs testing, lmk if you need a hand with that!

zloery:
We've confirmed this fix works on my integration environment for both previously-working and previously-broken users.

PR #177, opened 2020-10-23
Make legacy configs app generator work for apps that don't have any instance types

choover-broad:
Make legacy configs app generator work for Sam

### DIFF
--- a/charts/terra-argocd-app/Chart.yaml
+++ b/charts/terra-argocd-app/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: terra-argocd-app
 description: A Helm chart for generating ArgoCD resources for a Terra app in an environment
 type: application
-version: 0.1.2
+version: 0.1.3
 dependencies:
 - name: terra-argocd-templates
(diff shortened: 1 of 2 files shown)


charts/terra-argocd-app/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: terra-argocd-app
 description: A Helm chart for generating ArgoCD resources for a Terra app in an environment
 type: application
-version: 0.1.2
+version: 0.1.3
```
gmalkov, line 5: Don't need to bump version anymore, if you really want a patch version you can comment #patch in the merge commit

gmalkov (approved):

charts/terra-argocd-app/Chart.yaml
choover-broad: derp, sorry, keep forgetting, thx!

gmalkov: All good, it's NBD

PR #179, opened 2020-10-27
(No Ticket) Adjustments to duos helm chart container naming

mflinn-broad:
This pr adjusts the names of k8s containers from the DUOS services so that it is easily identifiable which service 
service the container is associated with in metadata﻿


rushtong (approved):
👍

PR #192, opened 2020-11-05
[no ticket] remove dryRun

Qi77Qi:
I think we feel good enough about resource validator now that we can enable it for real


rtitle (approved):
🚀

PR #195, opened 2020-11-09
DDO-660 Helm chart updates for Leonardo GKE poc

choover-broad:
Update Leonardo Helm chart to include support for the following:
* Deploy the Leonardo app in multiple configurations (backend and frontend). Currently Deployments mount configuration files [from firecloud-develop](https://github.com/broadinstitute/firecloud-develop/pull/2343) as Kubernetes secrets; this will no longer be necessary once Leo has been cut over to GKE everywhere
* Create Ingress for Leonardo with [container-native load balancing](https://cloud.google.com/kubernetes-engine/docs/concepts/container-native-load-balancing) enabled. The Ingress has the same SSL policies, certificates, and backend timeout as the existing Leonardo load balancer
  * The old Leo lb redirects traffic from http to https; the new Ingress does not serve any traffic over http. (I need to check with the Leo team to see if this is acceptable)
* Enable Prometheus JVM monitoring, following @mflinn-broad's examples from other charts

Corresponding helmfile PR: https://github.com/broadinstitute/terra-helmfile/pull/549


mflinn-broad (approved):
looks great to me! You @gmalkov and I are using slightly different names for a lot of values, particularly those associated with the ingress. We should probably standardize them across charts at some point to avoid confusion down the road.

PR #199, opened 2020-11-12
DDO-849 consent-gke-ingress

mflinn-broad:
This pr will convert consent from using a kubernetes internal loadbalancer for access to the application
to use GKE ingress as an external https loadbalancer﻿


rushtong (approved):
Looks reasonable to me 👍

choover-broad (approved):
Looks good! IMO it's probably worth setting a TLS policy now rather than later, otherwise it'll probably just show up in prod scans again later and require remediation:

https://broadworkbench.atlassian.net/browse/DDO-494
https://broadworkbench.atlassian.net/browse/DDO-480

PR #182, opened 2020-10-28
DDO-777 Sam k8s port

gmalkov:

PR #200, opened 2020-11-12
DDO-790 rawls k8s initial

mflinn-broad:
Initial pr for the rawls helm chart

### DIFF
--- a/charts/rawls/templates/_deployment.tpl
+++ b/charts/rawls/templates/_deployment.tpl
@@ -46,7 +46,9 @@ spec:
         emptyDir: {}
       - name: rawls-prometheusjmx-jar
         emptyDir: {}
-
+      - name: {{ $settings.name }}-cm
+        configMap:
+          name: {{ $settings.name }}-cm
       containers:
       - name: {{ $settings.name }}-app
         image: "gcr.io/broad-dsp-gcr-public/rawls:{{ $imageTag }}"
@@ -69,6 +71,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.name
+        - name: PROMETHEUS_ARGS
+              value: "-javaagent:/etc/prometheusjmx/prometheusjmx.jar=9090:/etc/rawls-cm/prometheusJmx.yaml"
         volumeMounts:
         - mountPath: /etc/rawls.conf
           subPath: rawls.conf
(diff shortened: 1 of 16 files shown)


charts/rawls/templates/_deployment.tpl
```diff
@@ -0,0 +1,186 @@
...
+            fieldRef:
+              fieldPath: metadata.name
+        - name: PROMETHEUS_ARGS
+          value: "-javaagent:/etc/prometheusjmx/prometheusjmx.jar=9090:/etc/rawls-cm/prometheusJmx.yaml"
+        command: ["/bin/bash"]
+        args:
+        - '-c'
+        - >-
+          java ${JAVA_OPTS}
+          {{- if .Values.prometheus.enabled }}
```
gmalkov, line 81: IMO if we are going to have this flag we should make the chart work without Prometheus if it is disabled. See [this commit](https://github.com/broadinstitute/terra-helm/pull/182/commits/1f02162a633c959fe7c484183254f74045dc11b2) for an example of the changes needed.

charts/rawls/values.yaml
```diff
@@ -0,0 +1,70 @@
...
+  # deploymentDefaults.serviceName -- What to call the Service
+  serviceName: null
+  # deploymentDefaults.serviceIP -- Static IP to use for the Service. If set, service will be of type LoadBalancer
+  serviceIP: null
+  # deploymentDefaults.legacyResourcePrefix -- What prefix to use to refer to secrets rendered from firecloud-develop
+  # @default deploymentDefaults.name
+  legacyResourcePrefix: null
+
+# A map of rawls deployments. In Terra's production environment,
+# rawls is deployed as 2 separate services (frontend and backend) with different configurations.
```
gmalkov, line 69: This isn't super clear. Is a map supposed to go here? Maybe give an example?

gmalkov (approved):

charts/rawls/values.yaml
mflinn-broad: ah I just copy pasted this from the cromwell chart and did a find and replace for rawls... but yeah good point it is a little unclear.

charts/rawls/templates/_deployment.tpl
mflinn-broad: I'm not sure this flag is necessary. I can't think of a situation where we'd want to set it to false off the top of my head. I actually went through and removed the flag, but I missed this one. 

I think if we ever do have a use case for this flag its easy enough to go in and add it.

charts/rawls/values.yaml
mflinn-broad: ah I remember this now, this was a section that was needed in the cromwell chart but not in rawls, forgot to delete the comment

PR #203, opened 2020-11-17
[PF-136]: Add Janior related configs in RBS 

yonghaoy:
Note that Janitor configs are also in vault. So the value has Janitor's vault path, and we will set the value in helmfile


charts/rbs/README.md
```diff
@@ -19,7 +19,7 @@ Chart for Resource Buffering Service
 | certManager.issuerKind | string | `"ClusterIssuer"` |  |
 | certManager.issuerName | string | `"cert-manager-letsencrypt-prod"` |  |
 | certManager.renewBefore | string | `"360h0m0s"` | When to renew the cert. Defaults to 15 days before expiry. |
-| configBasedAuthzEnabled | bool | `true` | Whether to use static file with admin users email for authorization. TODO(PF-81): Switch to use SAM instead of config file for user authZ. |
+| cleanupAfterHandout | bool | `false` | Whether to publish message to Janitor after resource is cleaned up. |
```
wchamber, line 22: after a resource is handed out.

charts/rbs/values.yaml
```diff
@@ -27,9 +27,11 @@ vault:
   enabled: true
   # vault.pathPrefix -- (string) Vault path prefix for secrets. Required if vault.enabled.
   pathPrefix:
-  # vault.adminUserFilePath -- (string) Vault path prefix for admin user list. Required if vault.enabled.
-  # Use the same users as admin for all env by default. Override if needed in helmfile repo.
-  adminUserFilePath: config/terra/rbs/common/iam
+  # vault.janitorConfigPath -- (string) Janitor related configs in Vault path prefix. Required if vault.enabled.
+  # Now those Janitor configs are generated by Terraform and output to Vault.
```
wchamber, line 31: let's improve this comment. I think this is also required if cleanupAfterHandout is true. Can we be more specific than `Janitor related configs`? Maybe this ssould be janitorPubSubConfigPath?

wchamber (approved):

gmalkov (changes requested):
This chart grabbing the config "secrets" for pubsub that Janitor makes didn't feel great to me, so I had a discussion for best practices around this kind of configuration with the rest of the AP DevOps team.

Since these charts are meant to be portable and eventually consumable by third parties, we don't want to bake in a dependency on Vault.

For actual secrets in Helm chart, our best practice is to allow them to be synced from vault, but also allow them to be populated in some other way. This is easy to do by enclosing the `SecretDefinition` resources in a conditional, since anyone not using Vault can see what the secrets being mounted are and populate them manually.

For configuration, like the PubSub details, this is more tricky to do. If we take a similar approach, optionally excluding the `SecretDefinition` resources, to anyone not familiar with our pattern of using Vault as a k/v store sourcing config from k8s secrets will seem very strange.

So our policy going forward will be that any values output by Terraform and needed by Helm that aren't secret should be output by the TF configuration in `terraform-ap-deployments` and manually entered in the corresponding envs' values in `terra-helmfile`. We realize this is inconvenient, but encouraging this awkward use of Vault to ferry configuration will result in a lot of extra work for us down the line when we will undoubtedly have to eventually rip it all out and replace it with a more k8s-natve way for managing configuration.

wchamber:
@gmalkov Thanks, I'll start a thread to discuss and we can move to a meeting as needed.
More generally, I think this is the kind of decision that would be very helpful to see in terms of a doc with pros/cons and alternatives considered. That would help us to centralize discussion and build reasoning that we can re-examine later.

gmalkov (approved):
We can merge this for now but let's re-visit the config piping from Terraform to k8s soon.

PR #208, opened 2020-11-23
DDO-835 Add slack to prometheus alertmanager

mflinn-broad:
Adding support for alert manager to fire to slack. see [here](https://github.com/broadinstitute/terra-helm/blob/7d4ce2461cd1479d62cdb15c84c2ea5106ce8a05/charts/terra-prometheus/templates/alertmanager-conf.yaml#L18) for more info on details of this change


gmalkov (approved):
That's annoying, great explanation write-up though!

PR #210, opened 2020-11-24
[PF-232] fix RBS image name

yonghaoy:
The github repo name is terra-resource-buffer:
https://github.com/DataBiosphere/terra-resource-buffer
And this should be the image name

### DIFF
--- a/charts/buffer/values.yaml
+++ b/charts/buffer/values.yaml
@@ -16,7 +16,7 @@ appVersion:
 
 imageConfig:
   # imageConfig.repository -- Image repository
-  repository: gcr.io/terra-kernel-k8s/terra-buffer
+  repository: gcr.io/terra-kernel-k8s/terra-resource-buffer
   # imageConfig.tag -- (string) Image tag.
   # @default -- The chart's appVersion value will be used
   tag:


charts/buffer/values.yaml
```diff
@@ -16,7 +16,7 @@ appVersion:
 
 imageConfig:
   # imageConfig.repository -- Image repository
-  repository: gcr.io/terra-kernel-k8s/terra-buffer
+  repository: gcr.io/terra-kernel-k8s/terra-resource-buffer
```
choover-broad, line 19: LGTM!

I have a slight preference for renaming the docker image to `buffer`, which matches the conventions the other Terra apps use (Cromwell is called `cromwell`, Leonardo is called `leonardo`, etc). I'm not sure why the image needs to be qualified with `terra-`, when it's already deployed to a repository that includes the name `terra`.

choover-broad (approved):

charts/buffer/values.yaml
yonghaoy: I think the convention is to use the github repo name:
https://github.com/DataBiosphere/terra-resource-buffer/blob/master/.github/workflows/master_push.yml#L74
And I think that is what janitor, WM and RBS did. And I think @gjuggler mentioned somewhere that the new services repo name should starts with `terra-`

PR #214, opened 2020-11-30
DDO-904 rawls ingress health checks

mflinn-broad:
This pr sets up a backend config for the rawls ingress so that the ingress
health check is configured correctly on pods with multiple containers﻿


choover-broad (approved):
LGTM!

nit: there's a `rawls-frontend-config` and `rawls-ingress-backendconfig`, should the naming convention match? (whatever you pick I can also update Leo to match)

mflinn-broad:
> LGTM!
> 
> nit: there's a `rawls-frontend-config` and `rawls-ingress-backendconfig`, should the naming convention match? (whatever you pick I can also update Leo to match)

Good point, the later is more clear  IMO. Shall we go with that?

PR #217, opened 2020-12-01
PF-153 Add WM chart values for spend and google folder for creating projects

wchamber:
This will enable us to configure WM with a spend profile id/billing account id and a google folder to create projects in.

While I'm not crazy about the continued proliferation of helm chart values & environment variables, we are still awaiting a long term solution for passing the application.yaml through the helm chart. So good enough for now.


charts/workspacemanager/templates/deployment.yaml
```diff
@@ -98,6 +98,12 @@ spec:
           value: "{{ .Values.cloudTraceEnabled }}"
         - name: TERRA_DATAREPO_URL
           value: "{{ .Values.terraDataRepoUrl }}"
+        - name: WORKSPACE_GOOGLE_FOLDERID
+          value: "{{ .Values.google.folderId }}"
+        - name: WORKSPACE_SPEND_SPENDPROFILES_0_ID
```
zloery, line 103: Why index in the env var names, but not the helm var names? Do we expect to use more than one spend profile/billing account ID?

zloery (approved):
LGTM + 1 naming nit

charts/workspacemanager/templates/deployment.yaml
wchamber: The spring configuration yaml setup we have allows us to have multiple spend profile ids. The env var names are mapping to that setup directly via Spring relaxed binding magic. https://github.com/spring-projects/spring-boot/wiki/Relaxed-Binding-2.0#environment-variables This allows us to avoid the intermediate env var -> spring applicatoin value mapping.
So the index in the helm chart name is because [this maps to a list](https://github.com/DataBiosphere/terra-workspace-manager/blob/cf6af10fea189bba6742e29a134ab66f81b3af20/src/main/java/bio/terra/workspace/app/configuration/external/SpendProfileConfiguration.java#L21).

I thought it was important in the WM code to allow multiple spend profile ids to better match eventual SPM. I don't like that our current helm value set up requires us to add a 2 helm chart values & 2 env variables for every supported spend profile id. In practice for the short term, I only expect to need one since we're having a single payer for the phase 1 pilot. So I didn't want to add more helm values than we need or mess around with dynamically iterating through arrays of objects to create arbitrarily many env variables. So there is no index for the helm var.

PR #221, opened 2020-12-03
Add locking to release workflows

gmalkov:
This avoids race conditions when writing to index.yaml while releasing charts

### DIFF
--- a/.github/workflows/pr-release.yaml
+++ b/.github/workflows/pr-release.yaml
@@ -6,6 +6,17 @@ jobs:
   release:
     runs-on: ubuntu-latest
     steps:
+      - name: Set up GCS Google Service Account
+        run: echo '${{ secrets.GCS_SA }}' > service_account.json
+
+      - name: Lock
+        uses: databiosphere/github-actions/actions/locker@locker-0.8.0
+        with:
+          operation: lock
+          lock_name: terra-helm-release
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-release.yaml
```diff
@@ -6,6 +6,17 @@ jobs:
...
+      - name: Set up GCS Google Service Account
+        run: echo '${{ secrets.GCS_SA }}' > service_account.json
+
+      - name: Lock
+        uses: databiosphere/github-actions/actions/locker@locker-0.8.0
+        with:
+          operation: lock
+          lock_name: terra-helm-release
+          bucket: broad-dsp-locker-action
+          lock_timeout_ms: '120000'
```
mflinn-broad, line 18: Does this mean the lock will release no matter what after 120 seconds or the action will fail after trying to aquire the lock for 120 seconds?

mflinn-broad (approved):

.github/workflows/pr-release.yaml
gmalkov: It will fail, there is another [optional parameter](https://github.com/DataBiosphere/github-actions/tree/master/actions/locker#inputs), `continue_on_lock_timeout` that can be set to true to make it continue instead of failing.

PR #222, opened 2020-12-04
DDO-905 Pull Agora cert from Vault

choover-broad:


choover-broad:
Closing - after discussion on Slack we're going to stick with the terraform-firecloud certs

PR #233, opened 2020-12-15
DDO-915 Automatically reset all Terra ArgoCD Apps back to master branch of terra-helmfile every night

choover-broad:
* Configure ArgoCD apps to deploy HEAD revision of terra-helmfile by default (previously this was not specified)
* Add a sync window + automated sync policy to terra-app-generator to automatically sync during a 10-minute window every night at 10pm (the nightly perf deploy kicks off at

### DIFF
--- a/charts/dsp-argocd/templates/apps/terra-app-generator.yaml
+++ b/charts/dsp-argocd/templates/apps/terra-app-generator.yaml
@@ -9,6 +9,11 @@ spec:
     server: https://kubernetes.default.svc # And in the same cluster
   sourceRepos:
   - '*'
+  syncWindows:
+  - kind: allow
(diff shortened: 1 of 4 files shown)


gmalkov:
Will this apply just to perf or to all apps in all envs? We don't want peronal env apps set to custom branches resetting every night, right?

choover-broad:
@gmalkov does anyone actually use ArgoCD to deploy to personal envs?

choover-broad:
It's all envs, I can add a config switch to exclude personal envs, though. Just means a bit more complexity in terra-helmfile

gmalkov:
> @gmalkov does anyone actually use ArgoCD to deploy to personal envs?

I do sometimes. Dunno if anyone else does. I guess we can encourage use of tools like Skaffold for that use case instead.

gmalkov (approved):

PR #234, opened 2020-12-15
DDO-945 First pass at cromiam Helm chart

choover-broad:
Initial cromiam Helm chart implementation

### DIFF
--- a/charts/cromiam/templates/configmap.yaml
+++ b/charts/cromiam/templates/configmap.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Chart.Name }}-cm
+  labels:
+{{ include "cromiam.labels" . | indent 4 }}
+data:
+  prometheusJmx.yaml: |
+{{ include "cromiam.config.prometheusJmx" . | indent 4 }}
--- a/charts/cromiam/templates/ingress.yaml
+++ b/charts/cromiam/templates/ingress.yaml
@@ -1,0 +1,19 @@
+{{ if .Values.ingress.enabled -}}
+apiVersion: networking.k8s.io/v1beta1
+kind: Ingress
+metadata:
+  name: {{ .Chart.Name }}-ingress
+  labels:
+{{ include "cromiam.labels" . | indent 4 }}
+  annotations:
+    networking.gke.io/v1beta1.FrontendConfig: "{{ .Chart.Name }}-ingress-frontendconfig"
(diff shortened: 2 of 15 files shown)


emandakh (approved):

charts/cromiam/templates/configmap.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Chart.Name }}-cm
```
mflinn-broad, line 4: I noticed Greg was using a  `{{ .Values.name }}` for this purpose and I think that is a good pattern. It makes it much easier if we need to make tweaks to application names or override them for some reason down the road.

charts/cromiam/templates/ingress.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  name: {{ .Chart.Name }}-ingress
+  labels:
+{{ include "cromiam.labels" . | indent 4 }}
+  annotations:
+    networking.gke.io/v1beta1.FrontendConfig: "{{ .Chart.Name }}-ingress-frontendconfig"
+    kubernetes.io/ingress.global-static-ip-name: {{ required "ingress.staticIpName value is required" .Values.ingress.staticIpName | quote }}
+{{- if not (empty .Values.ingress.preSharedCerts) }}
+    ingress.gcp.kubernetes.io/pre-shared-cert: {{ .Values.ingress.preSharedCerts | join "," | quote }}
+{{- end }}
```
mflinn-broad, line 13: If you aren't including the logic to pull the  certs from vault also then I don't think this conditional is necessary. Not a big deal though 😄

mflinn-broad (approved):
woohoo!

charts/cromiam/templates/ingress.yaml
choover-broad: Don't want to supply an empty string to the annotation and get potentially undefined behavior 🤷‍♀️

mflinn-broad: not trying to bike shed but would it be better to use the `required` function with the value in that case?

charts/cromiam/templates/configmap.yaml
choover-broad: Is the idea that anywhere .Chart.Name would be used, `.Values.name` is substituted? I.e., I shouldn't have any uses of `.Chart.Name` in the chart at all?

mflinn-broad: I think so @gmalkov any thoughts on this?

mflinn-broad: Also, didn't mean for these to be blockers to you merging the pr. I intended them more as general best practice questions 😄

PR #239, opened 2020-12-17
DDO-944 Move DUOS to use ingress instead of lb

rushtong:
Partially addresses https://broadworkbench.atlassian.net/browse/DDO-944


mflinn-broad (approved):
This all looks correct to me! I think we should try deploying from your pr branches just to be sure before merging anything though.

mflinn-broad (approved):
lgtm!

PR #240, opened 2020-12-17
[PF-197] Remove WM references to cloud trace SA

zloery:
Workspace Manager stopped using a separate service account for cloud tracing in https://github.com/DataBiosphere/terra-workspace-manager/pull/121, but the existing SA credentials were left in for backwards compatibility. 

All environments are now on a version past this change (see https://github.com/broadinstitute/terra-helmfile/pull/789), so it's now safe to remove this from the Helm chart. I'll make a terraform change to remove this SA once this chart change is deployed to all environments.

No helmfile change is necessary as there is no cloudtrace SA specific value, and we continue to use `Values.cloudTraceEnabled`.

### DIFF
--- a/charts/workspacemanager/templates/deployment.yaml
+++ b/charts/workspacemanager/templates/deployment.yaml
@@ -43,9 +43,6 @@ spec:
         secret:
           secretName: workspace-sqlproxy-sa
       {{- end }}
-      - name: cloud-trace-sa-creds
-        secret:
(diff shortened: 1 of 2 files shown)


zloery:
Tested using skaffold, WM deploys successfully to a personal environment

mflinn-broad (approved):
👍🏻

PR #241, opened 2020-12-18
don't send worspace manager logs to kibana to avoid quota issues

mflinn-broad:
This is my quick hack to  stop hammering our non prod logit stack. The workspace manager proxy has been sending a ton of logs. As a quick fix I set filebeat to not forward workspace manager logs to kibana. 

Will make a DDO ticket to come up with a better longterm solution. Perhaps a flag so that only legacy(former GCE) apps deployed on k8s forward to kibana


choover-broad (approved):
Awesome @mflinn-broad !!!

PR #247, opened 2021-01-04
DDO-963 Update default cert renewal interval

gmalkov:
Letsencrypt sends out warning emails 20 days before expiry, and our cert-manager certs for these charts are currently set to auto-renew 15 days before expiry. Changing that to 30 days to avoid the email spam.


choover-broad (approved):
Nice!

mflinn-broad (approved):
good call!

PR #248, opened 2021-01-04
DDO-919 Final part: Flag to disable cronjob alerts

mflinn-broad:
Leonardo's resource-validator cronjob consistently fails on every run in the perf environment. This pr adds a flag to the prometheus alerting rules so that we can optionally disable alerting on failed cronjobs for a particular environment. 

The default behavior is that alerting on failed cronjobs is enabled. It must be explicitly disabled for a particular environment.


charts/terra-prometheus/templates/rules/appRules.yaml
```diff
@@ -118,11 +118,13 @@ spec:
       for: 10m
       labels:
         severity: warning
+    {{- if .Values.enableCronjobAlerts }}
     - alert: KubeJobFailed
       annotations:
         message: Job {{`{{`}} $labels.namespace {{`}}`}}/{{`{{`}} $labels.job_name {{`}}`}} failed to complete.
         runbook_url: alert-name-kubejobfailed
       expr: kube_job_status_failed{job="kube-state-metrics"}  > 0
```
choover-broad, line 126: any way we can suppress alerting for just the broken Leo cron jobs? Wondering if it would be better to disable them in perf in the Leo chart instead of disabling alerting for all cron jobs in perf...

choover-broad (approved):

charts/terra-prometheus/templates/rules/appRules.yaml
mflinn-broad: Perhaps the best option here is to use labels to more specifically target the leo crons. Let me see if I can come up with a better approach

choover-broad (approved):
Awesome! 🎉

PR #255, opened 2021-01-12
Resetting opendj to empty chart so it will deploy

gmalkov:
The current chart causes errors when terra-helmfile tries to deploy it, so rather than debug them I am going to wipe it out since the new chart is written from scratch anyway.


choover-broad (approved):
💣

PR #258, opened 2021-01-13
DDO-970 elasticsearch POC

mflinn-broad:
[argo cd](https://ap-argocd.dsp-devops.broadinstitute.org/applications/elasticsearch-dev)


charts/elasticsearch/README.md
```diff
@@ -0,0 +1,71 @@
+# elasticsearch
```
rmeffan, line 1: Even viewing this document as source makes me so happy. Very nicely done.

charts/elasticsearch/templates/backup/cronjob.yaml
```diff
@@ -0,0 +1,36 @@
...
+          restartPolicy: OnFailure
+          nodeSelector: {{ .Values.backup.nodeSelector | toYaml | nindent 12 }}
+          tolerations: {{ .Values.backup.tolerations | toYaml | nindent 12 }}
+          containers:
+          - name: {{ .Values.name }}-backup
+            image: {{ .Values.backup.imageRepo}}:{{ .Values.backup.imageTag }}
+            command: ['sh', '-c']
+            args:
+            - |-
+              TIMESTAMP=$( date "+%Y%m%d.%H%M%S" ) && \
```
rmeffan, line 29: Unsure how this might be different for what looks like Bash inside a yaml file entry, but date format strings are usually single-quoted, and the `+` sign should appear before the opening quote (unless you intend it to be printed as a literal).

Example for your case: `+'%Y%m%d.%H%M%S'`

charts/elasticsearch/templates/backup/cronjob.yaml
```diff
@@ -0,0 +1,36 @@
...
+              curl -X PUT "${ELASTICSEARCH_HOSTNAME}:9200/_snapshot/${SNAPSHOT_REPOSITORY}/es-snapshot-${TIMESTAMP}?wait_for_completion=true"
```
rmeffan, line 30: If you append `:?` to your Bash variable name references, you'll get an error with a default message if the variable is null or unset.

charts/elasticsearch/values.yaml
```diff
@@ -1 +1,110 @@
...
+      volumes:
+        - secret
+        - configMap
+        - persistentVolumeClaim
+        - emptyDir
+
+  rbac:
+    create: true
+  
+  service:
```
rmeffan, line 55: Will this stanza create a LoadBalancer? Unless something has changed, our existing Terra Elasticsearch use case does not have an LB fronting it.

Note:
- This may be obvious.
- Assume you already know this.
- I don't know enough about Helm charts to know if this stanza will do anything.

rmeffan (commented):
Overall, LGTM. Can't speak to the Helm accuracy, but added some comments.

charts/elasticsearch/README.md
mflinn-broad: It's actually auto-generated documentation using [helm-docs](https://github.com/norwoodj/helm-docs)

charts/elasticsearch/templates/backup/cronjob.yaml
mflinn-broad: TIL

charts/elasticsearch/templates/backup/cronjob.yaml
mflinn-broad: TIL again 👍🏻

gmalkov (approved):

PR #264, opened 2021-01-20
DDO-975 opendj chart

gmalkov:


choover-broad (approved):
Looks great! Very clean, I especially like the separate template for firewall rules. A few thoughts:
* We might need to add additional vm IPs to terra-helmfile before cutting over DNS (eg. thurloe, job manager)
* Are there already backups in place for the existing persistent disk? (I'm currently trying to figure this out for MongoDB)

PR #266, opened 2021-01-21
[PF-326] Organize buffer variables into a single group and add enabled flag.

tnadler:


charts/workspacemanager/templates/secrets.yaml
```diff
@@ -151,4 +151,10 @@ spec:
       path: {{ .Values.vault.pathPrefix }}/workspace/app-sa
       encoding: base64
       key: key
-{{ end }}
+    {{- if .Values.buffer.enabled }}
+    buffer-client-sa.json:
+      path: {{ .Values.buffer.clientCredentialPath }}
```
gmalkov, line 156: Adding a [`required` wrapper](https://helm.sh/docs/howto/charts_tips_and_tricks/#know-your-template-functions) to this value will ensure that it will fail to render if `buffer.enabled` is `true` but no value for it is set, which is why our env-diff check didn't catch it before. It's valid YAML without a value here but not a valid `SecretDefinition`

gmalkov (approved):
LGTM, just a minor tweak

charts/workspacemanager/templates/secrets.yaml
tnadler: PTAL

gmalkov (approved):

PR #271, opened 2021-01-22
DUOS-954 - Consent/Ontology replicas value

rushtong:
Remove the default replicas so we pull it from the values file

Fallout from https://broadworkbench.atlassian.net/browse/DUOS-954


choover-broad (approved):
LGMT! Just want to note that it's fine to leave a sensible default value of `1` here and just let `3` in the helmfile repo override it if that's your preference!

rushtong:
Ok, I think I can kill this PR entirely then if the values here will be overridden.

PR #277, opened 2021-01-29
Add storage class name

gmalkov:


mflinn-broad (approved):
I wonder if it makes sense to just include the changes from [this pr](https://github.com/broadinstitute/terra-helm/pull/291) here too. I think both of these changes are going to force deleting and then creating new `persistentVolumeClaims` anyways.

PR #279, opened 2021-02-04
Sam label tweaks

gmalkov:


charts/sam/templates/configmap.yaml
```diff
@@ -3,9 +3,9 @@ kind: ConfigMap
 metadata:
   name: {{ .Values.name }}-configmap
   labels:
-{{ include "sam.labels" . | indent 4 }}
+    {{- include "sam.labels" . | nindent 4 }}
 data:
 {{- if .Values.prometheus.enabled }}
   prometheusJmx.yaml: |
-{{ include "sam.config.prometheusJmx" . | indent 4 }}
+    {{- include "sam.config.prometheusJmx" . | nindent 4 }}
```
mflinn-broad, line 10: for my own curiosity, what's the advantage of this syntax?

mflinn-broad (approved):

charts/sam/templates/configmap.yaml
gmalkov: it enables having nice indentation instead of always having to put the brackets all the way on the left

PR #283, opened 2021-02-07
Bump backend timeout for ArgoCD's Ingress

choover-broad:
The Jenkins deploy process uses diffs from ArgoCD to identify which applications need a sync. Lately it's been generating errors like this:

```
[ warn]  Failed to pull a diff for consent-prod from argocd, will retry 1 more times
[debug]  $ argocd app diff consent-prod --hard-refresh
time="2021-02-02T21:01:08Z" level=fatal msg="rpc error: code = Unknown desc = POST https://ap-argocd.dsp-devops.broadinstitute.org:443/application.ApplicationService/Get failed with status code 502"
```

I think this is because the diff operation is taking longer than 30s. If it can't generate a diff, the deploy process performs a sync anyway, but these messages are confusing and it would be nice to quiet them.

### DIFF
--- a/charts/dsp-argocd/templates/backendconfig.yaml
+++ b/charts/dsp-argocd/templates/backendconfig.yaml
@@ -8,5 +8,6 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     app.kubernetes.io/part-of: {{ .Chart.Name }}
 spec:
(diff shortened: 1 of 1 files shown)


mflinn-broad (approved):

choover-broad:
Applied manually via Helm

PR #284, opened 2021-02-07
DDO-992 Rework storage classes

choover-broad:
Replace old storage classes with new and improved ones:
* **`terra-zonal-ssd`**: SSD storage, `ext4` filesystem. A good default choice for zonal clusters.
* **`terra-zonal-xfs`**: SSD storage, `xfs` filesystem. Added to support MongoDB (the XFS filesystem is strongly [recommended]( http://dochub.mongodb.org/core/prodnotes-filesystem) for production MongoDB installations)
* **`terra-regional-ssd`** and **`terra-regional-xfs`**: [Regional](https://cloud.google.com/kubernetes-engine/docs/concepts/persistent-volumes#regional_persistent_disks) versions of the above, suitable for regional clusters like terra-prod.

Regional classes are configured [without a topology, which means they should replicate to a random availability zone in the same region](https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/regional-pd), allowing GKE to automatically launch replacement pods with the same data in the replicated zone in the event of an outage.

All storage classes now use the [`waitForFirstConsumer`](https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode) binding mode, which is [recommended](https://cloud.google.com/kubernetes-engine/docs/concepts/persistent-volumes#pd-zones) by GCP for dynamically-provisioned persistent disks.


gmalkov (approved):
Great write-up!

PR #285, opened 2021-02-09
Cleanup buffer.enabled and googleFolderId

tnadler:
Fields were removed in https://github.com/DataBiosphere/terra-workspace-manager/pull/215/files

### DIFF
--- a/charts/workspacemanager/values.yaml
+++ b/charts/workspacemanager/values.yaml
@@ -154,8 +154,6 @@ prometheus:
   jmxJarVersion: '0.13.0'
 
 buffer:
-  # buffer.enabled - When enabled, use Buffer service to create projects.
-  enabled: false
   # clientCredentialPath - the credentials to use for talking to buffer service. Required if enabled is true.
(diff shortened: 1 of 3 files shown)


charts/workspacemanager/values.yaml
```diff
@@ -171,9 +169,6 @@ buffer:
 # cloudTraceEnabled -- Whether to enable gcp cloud trace
 cloudTraceEnabled: true
 
-# googleFolderId -- the id of the Google Folder to create projects for workspaces within.
-googleFolderId:
```
gmalkov, line 175: The corresponding env var (`WORKSPACE_GOOGLE_FOLDERID`) should also be removed from the deployment definition, right?

gmalkov (approved):
Please [re-generate the README](https://github.com/broadinstitute/terra-helm#documentation) when changing values

charts/workspacemanager/values.yaml
tnadler: Yes, thank you

PR #286, opened 2021-02-10
[PF-161]: [Helm] Buffer multiple instance Stairway support

yonghaoy:


charts/buffer/templates/deployment.yaml
```diff
@@ -133,6 +133,15 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        # Environment variables for Kubernetes config
+        - name: BUFFER_KUBERNETES_POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: BUFFER_KUBERNETES_POD_NAME_FILTER
```
yonghaoy, line 141: @gmalkov Hi Greg, I am trying to use this filter to find the pod that host the app, and filter out other. pods(proxy stuff). From line 25, is `buffer-service` the right pod name to use?

Place to use the filter: https://github.com/DataBiosphere/terra-common-lib/blob/develop/src/main/java/bio/terra/common/kubernetes/KubeService.java#L103

Thanks

gjuggler (approved):

PR #287, opened 2021-02-10
DUOS: Log in JSON Format

rushtong:
Update the app configs to log in json format. This requires two PRs in the apps to support this logging, do not merge until these are merged/deployed:
* https://github.com/DataBiosphere/consent/pull/925
* https://github.com/DataBiosphere/consent-ontology/pull/386

Addresses
* https://broadworkbench.atlassian.net/browse/DUOS-791
* https://broadworkbench.atlassian.net/browse/DUOS-1028

### DIFF
--- a/charts/consent/Chart.yaml
+++ b/charts/consent/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: consent
-version: 0.7.0
+version: 0.7.1
 type: application
 description: A Helm chart for DUOS Consent
 keywords:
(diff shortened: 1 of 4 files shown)


charts/consent/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: consent
-version: 0.7.0
+version: 0.7.1
```
mflinn-broad, line 3: You don't need to bump the chart version numbers. This will be handled automatically when you merge the pr.

mflinn-broad (approved):
One small comment otherwise 👍🏻

charts/consent/Chart.yaml
rushtong: The action seems to not like it when these aren't bumped at the PR stage, leaving in for now and expecting that the final version will be different.

PR #288, opened 2021-02-10
DDO-1060 Disk-manager cronjob

mflinn-broad:


charts/diskmanager/templates/config/configMap.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.name }}-cm
+  labels:
+    {{ include "diskmanager.labels" . | nindent 4 }}
```
gmalkov, line 6:
```suggestion
    {{- include "diskmanager.labels" . | nindent 4 }}
```
Did you confirm this renders correctly? I think this will add two newlines, the one you have here and the one added by `nindent`. The `-` ensures any preceding whitespace is gobbled up before the newline is added.

charts/diskmanager/templates/secrets.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: SecretDefinition
+metadata:
+  name: {{ .Values.name }}-gcp-sa-secretdefinition
+  labels:
+    {{ include "diskmanager.labels" . | nindent 4 }}
+spec:
+  name: {{ .Values.name }}-gcp-sa
+  keysMap:
+    application_default_credentials.json:
+      path: {{ .Values.vault.pathPrefix }}/diskmanager/terraform/diskmanager-sa
```
gmalkov, line 11: I'd add a required condition to the pathPrefix, since this resource will be invalid if it is omitted

gmalkov (approved):

PR #307, opened 2021-03-03
[No Ticket] Create a Service Account for Elasticsearch backup Cronjob

mflinn-broad:
Last night 3/2/21 The elastic search backup cronjob failed to schedule. I realized I forgot to make a service account for it.
This pr creates a service account for the backup cron to use.﻿

### DIFF
--- a/charts/elasticsearch/templates/backup/role.yaml
+++ b/charts/elasticsearch/templates/backup/role.yaml
@@ -3,7 +3,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: {{ .Values.name }}-backup-role
-  labels: {{ include "elasticsearch.labels". | nindent 4 }}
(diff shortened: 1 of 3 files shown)


choover-broad (approved):
Any the cron job can't use the existing sa for elasticsearch?

mflinn-broad:
> Any  reason the cron job can't use the existing sa for elasticsearch?

I thought it might be better to make a new sa that we can configure more easily. The existing SA is created by the helm chart from elastic co. If having two different SA's is more confusing though, I can just use the existing one.

choover-broad:
> > Any  reason the cron job can't use the existing sa for elasticsearch?
> 
> I thought it might be better to make a new sa that we can configure more easily. The existing SA is created by the helm chart from elastic co and doesn't offer much control over it. If having two different SA's is more confusing though, I can just use the existing one.

Gotcha! Understood, in that case having a separate SA definitely makes sense!

PR #308, opened 2021-03-04
DDO-1102 Alert if cronjob does not run in 24 hour period

mflinn-broad:
This pr configures an alert which looks at the last time each cronJob was successfully scheduled and compares to the current time. If greater than 24 hours have elapsed since the last successful scheduling, an alert will fire. There is a 2 hour buffer before the alert fires to accommodate for transient timing issues and avoid potential alert spam since most of our cronjobs are daily.﻿

### DIFF
--- a/charts/terra-prometheus/templates/rules/appRules.yaml
+++ b/charts/terra-prometheus/templates/rules/appRules.yaml
@@ -126,3 +126,11 @@ spec:
       for: 1h
       labels:
         severity: warning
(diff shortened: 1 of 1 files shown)


mflinn-broad:
Screen Shot 2021-03-04 at 1 01 29 PM
Confirmed this creates the desired alert in prometheus ui

gmalkov (approved):

choover-broad (approved):
Niiiice! 💯

PR #322, opened 2021-03-09
Add opendj restore scripts

gmalkov:


mflinn-broad (approved):
looks good to me. not sure if a `.helmignore` needs to be updated here so Argo doesn't try to render these. Not sure if `charts/*/scripts` is ignored already or not

gmalkov:
> looks good to me. not sure if a `.helmignore` needs to be updated here so Argo doesn't try to render these. Not sure if `charts/*/scripts` is ignored already or not

I assumed it would only render the stuff in the templates folder but I'll add a .helmignore just in case, thanks for the tip!

PR #323, opened 2021-03-10
DDO-1132 Fix Tcell config for Buffer

mflinn-broad:
This pr includes fixes so that buffer's apache proxy is able to be properly monitored by tcell, one of the blockers for RBS going to prod.
Changes:
1. Update buffer's proxy to use the latest image `broadinstitute/openidc-proxy:tcell_3_1_0`. This is identical to the previous proxy image other than that it is using the latest version of tcell.
2. Rename the environment variables containing the tcell api key and app id. The latest version of tcell can read these values directly from the environment rather than trying to inject them into the config file as bash variables, which doesn't actually work. [more info](https://docs.tcell.io/docs/server-agent-options#section-environment-variables-and-config-file-properties)

### DIFF
--- a/charts/buffer/templates/configmaps/proxy-tcell-configmap.yaml
+++ b/charts/buffer/templates/configmaps/proxy-tcell-configmap.yaml
@@ -11,8 +11,6 @@ data:
       "version": 1,
       "applications": [
         {
-          "app_id": "$(TCELL_APP_ID)",
-          "api_key": "$(TCELL_API_KEY)",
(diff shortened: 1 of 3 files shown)


gmalkov (approved):

mflinn-broad:
@yonghaoy FYI this should take care of one of your blockers for RBS production readiness. https://broadworkbench.atlassian.net/browse/DDO-1132

yonghaoy (approved):
Thanks!

PR #329, opened 2021-03-17
DDO-1147 Add 3-hour startupProbe to Leo

choover-broad:
Currently, we've configured Leo with a 5-minute liveness probe. This means that if Leo's `/version` endpoint ever stops responding, after 5 minutes Kubernetes will restart the pod, saving the on-call engineer from being paged and having to restart the pod manually.

This is useful, but poses a problem for Liquibase migrations. Leo's `/version` endpoint does not start returning 200s until Liquibase migrations have finished; in the event a migration takes longer than 5 minutes, Kubernetes will kill the pod and potentially leave the database in a half-migrated state. (See [this thread](https://broadinstitute.slack.com/archives/C01E2DDTGQG/p1615304704000800) from a few weeks ago where the first statement of a migration was executed, but not the second, and I had to manually fix up the database to get Leo to start up.)

This PR adds a 3-hour [startupProbe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/), also on the /version endpoint. This way:
* we still get auto-restart behavior from Kubernetes liveness probes
* database migrations will only be interrupted if they exceed 3 hours

Props to @gmalkov for the elegant probe config pattern!

### DIFF
--- a/charts/leonardo/README.md
+++ b/charts/leonardo/README.md
@@ -12,16 +12,18 @@ A Helm chart for Leonardo, Terra's Jupyter notebook integration service
 | commonCronjob.reportDestinationBucket | string | `nil` |  |
 | cronjob.googleProject | string | `nil` |  |
 | cronjob.imageRepository | string | `"us.gcr.io/broad-dsp-gcr-public/resource-validator"` |  |
-| cronjob.imageTag | string | `"ed126d1"` |  |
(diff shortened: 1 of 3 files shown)


mflinn-broad (approved):

Qi77Qi (approved):
not sure exactly if I understand the diff very well..but the promise described in the description sounds great :)

PR #331, opened 2021-03-18
DUOS-1124 Move admin endpoints to private port

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1124

Move the exposed admin endpoint to a different port, `8081`.

TODO: ~Ensure that the new port is not exposed outside the cluster.~ Done.

### DIFF
--- a/charts/consent/Chart.yaml
+++ b/charts/consent/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: consent
-version: 0.10.0
+version: 0.10.1
 type: application
 description: A Helm chart for DUOS Consent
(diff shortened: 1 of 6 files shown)


mflinn-broad (approved):
I believe this should  be sufficient to do the trick of not exposing the admin. Approving but I would like to try deploying this change to dev to verify it has the intended effect before merging if you haven't done so already.

rushtong:
Have confirmed with argo deploys to dev that the /admin endpoint is now a 404 publicly.

mflinn-broad (approved):

PR #341, opened 2021-03-26
DDO-1108 Initial ncbiaccess chart

mflinn-broad:
At this point this chart has been tested in every env but prod. Due to unrelated reason ncbiaccess is broken in every env but prod. So this is as good as it's going to get for a first draft.


charts/ncbiaccess/templates/imagePullSecret.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: secrets-manager.tuenti.io/v1alpha1
+kind: SecretDefinition
+metadata:
+  name: {{ .Values.name }}-dockerhub-creds
+  labels:
+{{ include "ncbiaccess.labels" . | indent 4 }}
+spec:
+  name: {{ .Values.name }}-dockerhub-creds
+  type: kubernetes.io/dockerconfigjson
```
choover-broad, line 9: This is really interesting, props for figuring it out!

choover-broad (approved):

charts/ncbiaccess/templates/imagePullSecret.yaml
mflinn-broad: NP. I'm going to do a write up in our gitbook of what is actually needed in order to pull from private repos. There are definitely some gotchas and conflicting information out there.

PR #360, opened 2021-04-09
[no ticket] fix hash bump

Qi77Qi:
https://github.com/broadinstitute/terra-helm/pull/359 worked as intended. But somehow argo parsed the hash as `              image: 'us.gcr.io/broad-dsp-gcr-public/zombie-monitor:5.186276e+06'`. Adding double quote and see if this fixes things


Qi77Qi:
going to merge since the change itself is pretty trivial and it's currently broken in dev for leo

choover-broad:
thanks for the fix @Qi77Qi! You can add `--tag '!!str'` to the yq command that updates these values to force quoting on ambiguous values like this

PR #361, opened 2021-04-12
*DDO-1176 add startup probes

knapii-developments:
Deployments failing due to long liquibase migrations have become a frequent occurrence.


charts/firecloudorch/templates/deployment.yaml
```diff
@@ -157,6 +157,9 @@ spec:
         livenessProbe:
           {{- toYaml .Values.probes.liveness.spec | nindent 10 }}
```
knapii-developments, line 158: @choover-broad I haven't seen this in any file. I just wanted to make sure that adding the start probe doesn't affect this.

choover-broad: Hey @knapii-developments, adding a separate startupProbe should be fine for `firecloudorch`! It has two possible liveness probes which are enabled depending on what values are supplied.

charts/firecloudorch/templates/deployment.yaml
```diff
@@ -157,6 +157,9 @@ spec:
...
         {{- end }}
+        startProbe:
```
choover-broad, line 160: spotted a typo here: should be `startupProbe`

charts/firecloudorch/templates/deployment.yaml
```diff
@@ -157,6 +157,9 @@ spec:
...
+          {{- toYaml .Values.probes.liveness.spec | nindent 10 }}
```
choover-broad, line 161: values here should be `.Values.probes.startup.spec`

knapii-developments: c1eed21

charts/firecloudorch/templates/deployment.yaml
knapii-developments: c1eed21

charts/sam/values.yaml
```diff
@@ -53,7 +53,17 @@ probes:
       periodSeconds: 10
       failureThreshold: 30 # 5 minutes before restarted
       successThreshold: 1
-
+  startup:
+    # probes.startup.enable -- Whether to configure a startup probe
+    enabled: false
```
knapii-developments, line 58: @choover-broad I noticed that the liveness probe wasn't enabled here. What are your thoughts?

charts/rawls/values.yaml
```diff
@@ -81,3 +81,14 @@ deploymentDefaults:
         periodSeconds: 10
         failureThreshold: 30 # 5 minutes before restarted
         successThreshold: 1
+    startup:
```
knapii-developments, line 84: @choover-broad This file has probes define. However, there's no mention of them in the [deployment file](https://github.com/broadinstitute/terra-helm/blob/master/charts/rawls/templates/deployments.yaml).

charts/sam/values.yaml
choover-broad: I believe it was disabled intentionally in order to prevent interruptions to liquibase migrations! So, I think we should enable it in this PR, too.

charts/rawls/values.yaml
knapii-developments: b54e7e2

charts/sam/values.yaml
choover-broad: argh, I think we'll need to revert the changes for ontology and firecloud orchestration as well, as they don't actually use liquibase. Apologies!

charts/firecloudorch/templates/deployment.yaml
choover-broad: Actually, it looks like firecloudorch won't need a startupProbe, as it does not use liquibase. Sorry for the mixed messages @knapii-developments!

charts/ontology/values.yaml
```diff
@@ -51,6 +51,17 @@ probes:
       initialDelaySeconds: 60
       periodSeconds: 10
       failureThreshold: 30 # 5 minutes before restarted
+  startup:
```
choover-broad, line 54: another chart that doesn't actually need a startupProbe, sorry! This one should be rolled back as well.

charts/cromiam/values.yaml
```diff
@@ -48,7 +48,17 @@ probes:
       periodSeconds: 10
       failureThreshold: 30 # 5 minutes before restarted
       successThreshold: 1
-
+  startup:
```
choover-broad, line 51: Ah, this is broken. It looks like [this change](https://github.com/broadinstitute/terra-helm/pull/297/files) from a few months ago broke cromiam's probe config, but the change was actually never deployed.

Let's correct `/version` to `/engine/v1/version` and `/status` to `/engine/v1/status` for all the cromiam probes.

charts/sam/values.yaml
knapii-developments: e732039

charts/firecloudorch/templates/deployment.yaml
knapii-developments: e732039

charts/cromiam/values.yaml
knapii-developments: a75d7d9

charts/sam/values.yaml
knapii-developments: dc66690

charts/ontology/values.yaml
knapii-developments: e732039

charts/buffer/templates/deployment.yaml
```diff
@@ -224,4 +224,13 @@ spec:
           name: tcell-config
           subPath: tcell-config.conf
         {{- end }}
+        {{- if .Values.probes.liveness.enabled }}
```
knapii-developments, line 227: @yonghaoy I added the probes here. This is 1/2 areas you need to review

charts/buffer/values.yaml
```diff
@@ -49,6 +49,29 @@ probes:
       periodSeconds: 10
       failureThreshold: 6
       successThreshold: 1
+    liveness:
```
knapii-developments, line 52: @yonghaoy I added the probe here. This is 2/2 areas you need to review.

charts/cromiam/values.yaml
```diff
@@ -29,7 +29,7 @@ probes:
     enabled: true
     spec:
       httpGet:
-        path: /status
+        path: /engine/latest/
```
choover-broad, line 32: `/engine/latest/status` 👍

choover-broad (approved):

PR #362, opened 2021-04-13
DUOS-1183 Update json logging format

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1183

## Changes
* Relabeled `level` as `severity` per Google documentation: https://cloud.google.com/logging/docs/structured-logging#special-payload-fields
* Added another service name field to simplify log filtering in Google's console

Example log:
```
{
  "severity":"INFO",
  "service-name":"consent",
  "thread":"main",
  "message":"Started @12509ms",
  "timestamp":1618328703081,
  "logger":"org.eclipse.jetty.server.Server"
}
```

## Testing
Manually tested with a deploy via Argo and triggered a 500 in: 

### Ontology:
Now shows the log message with the right severity level so we can easily filter on them:
Screen Shot 2021-04-13 at 3 18 19 PM

### Consent
Consent now logs with the correct severity in the test instances:
Screen Shot 2021-04-13 at 3 31 58 PM

### DIFF
--- a/charts/consent/Chart.yaml
+++ b/charts/consent/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: consent
-version: 0.13.0
+version: 0.12.0
 type: application
 description: A Helm chart for DUOS Consent
 keywords:
--- a/charts/consent/Chart.yaml
+++ b/charts/consent/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: consent
-version: 0.12.0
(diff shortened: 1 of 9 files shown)


mflinn-broad (approved):
everything looks good to me from the helm side of things. I haven't actually looked at the effect this has on the logging output in cloud logging.

rushtong:
@mflinn-broad I'm in testing mode today ... will make sure this logs as I expect it to.

PR #366, opened 2021-04-15
Bump resources on ArgoCD repo server to see if it helps address ArgoCD timeouts & slowness

choover-broad:
Bump requests to 4 Gib RAM / 4 CPUs for both repo server and controller containers, which seem to be resource constrained based on graphs in the GCP console. I'm skipping limits for now because I want to leave room for the containers to grow if needed.


rmeffan (approved):
LGTM

PR #372, opened 2021-05-12
DDO-1252 Adding prometheus-to-sd deployment for Kube State Metrics

mflinn-broad:


charts/terra-prometheus/values.yaml
```diff
@@ -41,7 +41,19 @@ vaultCert:
     # vaultCert.chain.secretKey -- (string) Key in secret containing intermediate .crt
     secretKey: 
 
-
+# prometheus-to-sd values. Used specifically for sending kube state metrics to sd
+prometheusToSD:
+  # prometheusToSD.imageRepo -- (string) image repo for prometheus-to-sd-exporter
+  imageRepo: gcr.io/google-containers/prometheus-to-sd
+  # prometheusToSD.imageRepo -- (string) image tag for prometheus-to-sd-exporter
```
choover-broad, line 48: typo, should be `imageTag`?

choover-broad (approved):
Awesome! 🎉 

Wondering if `deployment.yaml` and `serviceAccount.yaml` should be called something more specific or live under a subdirectory? Eg. `prometheus-to-sd/deployment.yaml`?

charts/terra-prometheus/values.yaml
mflinn-broad: yep, thanks

mflinn-broad:
> Awesome! 🎉
> 
> Wondering if `deployment.yaml` and `serviceAccount.yaml` should be called something more specific or live under a subdirectory? Eg. `prometheus-to-sd/deployment.yaml`?

Good call, will do!

PR #379, opened 2021-05-20
[IA-2228] Add Janitor cron job for Leonardo

kyuksel:
[JIRA ticket](https://broadworkbench.atlassian.net/browse/IA-2228)
[Companion leonardo-cron-jobs PR](https://github.com/broadinstitute/leonardo-cron-jobs/pull/91)

### DIFF
--- a/charts/leonardo/Chart.yaml
+++ b/charts/leonardo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: leonardo
 version: 0.38.0
 type: application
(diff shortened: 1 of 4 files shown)


Qi77Qi:
think you need to add something like [this](https://github.com/broadinstitute/terra-helm/blob/master/charts/leonardo/values.yaml#L101-L105) too?

would be nice to have this tested before merging

kyuksel:
> think you need to add something like this too?
Thanks @Qi77Qi for point that out. Pushed a commit to add it.

> would be nice to have this tested before merging
Certainly! Let's pair on that :)

PR #381, opened 2021-05-27
DDO-1268 Define azure credentials for atlantis declaratively

mflinn-broad:
This pr adds support for defining secrets holding credentials declaratively via secrets-manager rather than creating them manually. This pr adds support just for managing azure credentials but could easily be extended to support gcp credentials as well. 

Additionally we were previously using an atlantis chart from the now deprecated `helm-stable` chart repo which was deprecated  back in November. I switched the atlantis dependency over to the new atlantis chart repo so that we can actually build new versions of this chart.

### DIFF
--- a/charts/dsp-atlantis/templates/azureCredentials.yaml
+++ b/charts/dsp-atlantis/templates/azureCredentials.yaml
@@ -1,0 +1,34 @@
+{{- if .Values.azure.enabled -}}
+apiVersion: secrets-manager.tuenti.io/v1alpha1
+kind: SecretDefinition
+metadata:
+  name: {{ .Values.name }}-azure-creds-secretdefinition
+  labels:
+    {{- template "dsp-atlantis.labels" }}
+spec:
+  name: azure-cert
+  keysMap:
+    {{ .Values.azure.certificateFile }}:
+      path: {{ required "A valid .Values.vault.azure.pathPrefix is required" .Values.vault.azure.pathPrefix }}/certificate
+      encoding: base64
+      key: {{ .Values.azure.certificateFile }}.b64
+---
+apiVersion: secrets-manager.tuenti.io/v1alpha1
+kind: SecretDefinition
+metadata:
+  name: {{ .Values.name }}-azure-env-secretdefinition
+  labels:
+    {{- template "dsp-atlantis.labels" }}
+spec:
+  name: azure-env
+  keysMap:
+    ARM_CLIENT_CERTIFICATE_PASSWORD:
+      path: {{ required "A valid .Values.vault.azure.pathPrefix is required" .Values.vault.azure.pathPrefix }}/atlantis/env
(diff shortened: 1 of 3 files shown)


charts/dsp-atlantis/templates/azureCredentials.yaml
```diff
@@ -0,0 +1,34 @@
...
+  keysMap:
+    ARM_CLIENT_CERTIFICATE_PASSWORD:
+      path: {{ required "A valid .Values.vault.azure.pathPrefix is required" .Values.vault.azure.pathPrefix }}/atlantis/env
+      key: ARM_CLIENT_CERTIFICATE_PASSWORD
+    ARM_CLIENT_ID:
+      path: {{ required "A valid .Values.vault.azure.pathPrefix is required" .Values.vault.azure.pathPrefix }}/atlantis/env
+      key: ARM_CLIENT_ID
+    ARM_TENANT_ID:
+      path: {{ required "A valid .Values.vault.azure.pathPrefix is required" .Values.vault.azure.pathPrefix }}/atlantis/env
+      key: ARM_TENANT_ID
```
jack-r-warren, line 33: Does this replace [https://github.com/broadinstitute/terra-helm-definitions/blob/master/dsp-tools/dsp-atlantis/values.yaml#L46](https://github.com/broadinstitute/terra-helm-definitions/blob/master/dsp-tools/dsp-atlantis/values.yaml#L46) or is it creating the secrets that file references?

charts/dsp-atlantis/values.yaml
```diff
@@ -12,3 +13,9 @@ atlantis:
   fullnameOverride: atlantis
   serviceAccount:
     name: atlantis
+azure:
+  enabled: false
+  certificateFile: "azure-cert.pfx"
+vault:
+  azure:
+    pathPrefix: /path/to/azure/secrets
```
jack-r-warren, line 21: This seems like a dummy value, will anything that cares attempt to use it?

jack-r-warren (approved):
Just a few (selfish) questions on how this fits into how atlantis is configured, but generally looks good to me!

charts/dsp-atlantis/templates/azureCredentials.yaml
mflinn-broad: It creates the secrets referenced in that file

mflinn-broad: rather than me creating them manually

charts/dsp-atlantis/values.yaml
mflinn-broad: It is a dummy value it gets overriden in [this pr](https://github.com/broadinstitute/terra-helm-definitions/pull/53). 
That is the only place these values are used in our setup. 

When we auto generate documentation for this chart that dummy value will show up as in example which imo is better for documentation than leaving the real vault path in there.

PR #387, opened 2021-06-07
[DDO-1292] Rawls Liquibase Migration Pre-sync Job

jack-r-warren:
Summary of [DDO-1292](https://broadworkbench.atlassian.net/browse/DDO-1292):
- New liquibase.properties file produced by firecloud-develop ([DDO-1297](https://broadworkbench.atlassian.net/browse/DDO-1297))
  - Rationale: Rawls's DB config is in rawls.conf produced by firecloud-develop, but we can't read that from CLI
  - PR: https://github.com/broadinstitute/firecloud-develop/pull/2565
  - Currently deployed to [dev](https://ap-argocd.dsp-devops.broadinstitute.org/applications/rawls-configs-dev)
  - In the future, DB config would be helm-templated from single source of truth
- K8s job (and related resources) are applied before rest of Rawls in ArgoCD sync, running Liquibase via the app's image before the app ever starts
  - Rationale: Now the app doesn't need to run Liquibase migrations on startup--the job runs to completion and kills its own CloudSQL proxy sidecar to let the ArgoCD sync continue ([DDO-1301](https://broadworkbench.atlassian.net/browse/DDO-1301))
- Values file enabling this lives over in terra-helmfile, that's where we'd make it do more than a dry-run
  - PR: https://github.com/broadinstitute/terra-helmfile/pull/1349
  - After this merge that PR will be updated to refer to released helm chart
  - Currently deployed to [dev](https://ap-argocd.dsp-devops.broadinstitute.org/applications/rawls-dev)


charts/rawls/templates/migration/job.yaml
```diff
@@ -0,0 +1,88 @@
...
+{{- $imageTag := .Values.migration.imageTag | default .Values.global.applicationVersion -}}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: rawls-liquibase-migration
+  annotations:
+    argocd.argoproj.io/hook: PreSync
+    argocr.argoproj.io/hook-delete-policy: BeforeHookCreation
+  labels: 
+    {{- include "rawls.labels" . | nindent 4 }}
```
mflinn-broad, line 11: I realize rawls doesn't do this in other places but when you use `nindent` you can actually define labels like this. It reads a little cleaner this way imo.
```suggestion
  labels: {{- include "rawls.labels" . | nindent 4 }}
```

charts/rawls/templates/migration/job.yaml
```diff
@@ -0,0 +1,88 @@
...
+            name: {{ .Values.migration.secretPrefix }}-sqlproxy-env
+        volumeMounts:
+        - mountPath: /etc/sqlproxy-service-account.json
+          subPath: sqlproxy-service-account.json
+          name: sqlproxy-ctmpls
+          readOnly: true
+        command: ['sh', '-c']
+        args:
+        - |-
+          /cloud_sql_proxy ${CLOUDSQL_LOGGING:-"-verbose"} \
```
mflinn-broad, line 84: do we need verbose logging? Or is that just how it's set on our other proxies?

charts/rawls/templates/migration/role.yaml
```diff
@@ -0,0 +1,18 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: rawls-migration-role
+  labels:
+    {{- include "rawls.labels" . | nindent 4 }}
+  annotations:
+    argocd.argoproj.io/hook: PreSync
+    argocr.argoproj.io/hook-delete-policy: BeforeHookCreation
+    argocd.argoproj.io/sync-wave: "-1"
```
mflinn-broad, line 11: Is this to ensure the rbac resources get created before the acutal job?

I actually like creating and destroying these on each run of the hook. Helps reduce clutter in the argo ui.

mflinn-broad (approved):
Awesome work @jack-r-warren! looks great overall. Just a few small questions and nits.

charts/rawls/templates/migration/job.yaml
jack-r-warren: That's how it is set on all other proxies. I basically reduced the entire 160 line script down to that shell command, minus the support for not passing any credentials (we'd just omit the credentials line) and the "you didn't pass a required variable" checks.

charts/rawls/templates/migration/job.yaml
```diff
@@ -0,0 +1,88 @@
...
+        - sqlproxy
+      volumes:
+      - name: app-ctmpls
+        secret:
+          secretName: {{ .Values.migration.secretPrefix }}-app-ctmpls
+      - name: sqlproxy-ctmpls
+        secret:
+          secretName: {{ .Values.migration.secretPrefix }}-sqlproxy-ctmpls
+      containers:
+      - name: migration-liquibase
```
choover-broad, line 34: Can we prefix the container name with `rawls`? Makes it easier to search for logs in Stackdriver!

charts/rawls/templates/migration/role.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.migration.enabled }}
```
choover-broad, line 1: Do we need separate RBAC resources for this job? I think the permissions will be very similar to whatever the Rawls application uses. (Apologies if there's already been some background discussion here that I missed!)

Seems like we could add PreSync to the existing resources to make sure they exist when the job is run, but not delete them before every sync?

charts/rawls/templates/migration/role.yaml
jack-r-warren: > Is this to ensure the rbac resources get created before the acutal job?
Yeah. Actually testing right now to see if the "HookSucceeded" deletion policy would work for us, leaving the resources around until the job was done.

choover-broad (approved):
This is a seriously fantastic first Helm PR @jack-r-warren! 💯🥇 

(Left a few comments inline, it may be that I'm just missing context)

charts/rawls/templates/migration/role.yaml
jack-r-warren: Believe it or not, it looks like the "HookSucceeded" policy secretly isn't supported on non-job resources. Literally a no-op. So unfortunately it looks like we might live with the clutter.

mflinn-broad: In that case I'm pro Chelsea's suggestion of just reusing the existing rawls RBAC resources.

charts/rawls/templates/migration/role.yaml
jack-r-warren: We don't strictly need separate RBAC, but we do need them to exist, and unfortunately: https://github.com/argoproj/argo-cd/issues/3502#issuecomment-620808412

> Argo CD hooks are not considered part of Argo CD application resources, and features like drift detection and health assessments do not work. So the solution should involve not having to resort to adding `"helm.sh/hook": pre-install` to a Namespace resource.

Despite the clutter that this involves (https://github.com/broadinstitute/terra-helm/pull/387#discussion_r651287596) I don't know of a better solution

charts/rawls/templates/migration/role.yaml
jack-r-warren: I don't know if we want to sacrifice health checks and stuff on the actual Rawls RBAC stuff: https://github.com/broadinstitute/terra-helm/pull/387#discussion_r651290417

charts/rawls/templates/migration/role.yaml
choover-broad: Does this comment offer a viable workaround? https://github.com/argoproj/argo-cd/issues/3502#issuecomment-620832168

I.e.
* RBAC resources sync in wave -2
* Job syncs in wave -1 with a Sync hook
* Everything else syncs in default wave 0

If that won't work, I'm fine with duplicating the RBAC resources!

jack-r-warren: I hadn't seen that, that's a good idea. I'd talked to Mike and just now put the RBAC stuff behind a flag so the job can reuse the existing one. I think the only downside to leaning on sync waves is that the normal rbac resources would need to have that sort of migration-specific attribute, but that's a small downside imo. Let's chat at standup

jack-r-warren: This worked!

choover-broad (approved):
Nice, looks great to me!

The values allow us to enable this in `dev` only for an arbitrary length of time, right?

jack-r-warren:
> Nice, looks great to me!
> 
> The values allow us to enable this in `dev` only for an arbitrary length of time, right?

Correct! Within each environment the values can control whether it is enabled (default false), whether it is dry run (default true), and whether failures will fail the sync (default true)

PR #391, opened 2021-06-14
IA-2818 add enable flag for all cron jobs

Qi77Qi:
* add `enable` flag for all cron jobs
* reduce janitor frequency

Need https://github.com/broadinstitute/terra-helmfile/pull/1364 to go in first I think

### DIFF
--- a/charts/leonardo/values.yaml
+++ b/charts/leonardo/values.yaml
@@ -98,16 +98,19 @@ cronjob:
   imageRepository: us.gcr.io/broad-dsp-gcr-public/resource-validator
   imageTag: 2ebc9e3
   googleProject:
+  enabled:
 janitorCron:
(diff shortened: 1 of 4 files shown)


xuscrbw (approved):

charts/leonardo/values.yaml
```diff
@@ -98,16 +98,19 @@ cronjob:
   imageRepository: us.gcr.io/broad-dsp-gcr-public/resource-validator
   imageTag: 2ebc9e3
   googleProject:
+  enabled:
```
choover-broad, line 101: use an explicit `false` here as a default instead of the implied `nil`?

choover-broad (approved):
LGTM, left a nit inline :shipit:

PR #394, opened 2021-06-15
[No Ticket] Support Easily disabling log forwarding to kibana for select services

mflinn-broad:
This pr adds a template to filebeat chart which can easily render simple regex's to selectively ignore logs from a list of services. This is to support starting to deprecate kibana in lower environments.


jack-r-warren (approved):
LGTM

choover-broad:
> @choover-broad choover-broad requested a review from jack-r-warren now

Oops!

choover-broad (approved):
Thanks for this @mflinn-broad!

PR #401, opened 2021-06-22
DDO-1304 added githubCredentials.yaml to import github

jyang-broad:
token from vault into atlantis

### DIFF
--- a/charts/dsp-atlantis/templates/githubCredentials.yaml
+++ b/charts/dsp-atlantis/templates/githubCredentials.yaml
@@ -1,8 +1,6 @@
-{{- /*
-  Credentials for github accounts to be extracted from Vault
-  Primary github account should be that of dsp-atlantis https://github.com/dsp-atlantis
-  Documentation for this api here: https://github.com/tuenti/secrets-manager
-*/}}
+# Credentials for github accounts to be extracted from Vault
+# Primary github account should be that of dsp-atlantis https://github.com/dsp-atlantis
+# Documentation for this api here: https://github.com/tuenti/secrets-manager
 
 apiVersion: secrets-manager.tuenti.io/v1alpha1
 kind: SecretDefinition
--- a/charts/dsp-atlantis/templates/githubCredentials.yaml
+++ b/charts/dsp-atlantis/templates/githubCredentials.yaml
@@ -19,4 +19,4 @@ spec:
   keysMap:
     dsp-atlantis-github-token:
       path: {{ required "A valid .Values.vault.github.pathPrefix is required" .Values.vault.github.pathPrefix }}/tokens
-      key: ATLANTIS_GITHUB_TOKEN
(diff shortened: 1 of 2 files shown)


charts/dsp-atlantis/values.yaml
```diff
@@ -20,6 +20,8 @@ azure:
 vault:
   azure:
     pathPrefix: /path/to/azure/secrets
+  github:
+    pathPrefix: /path/to/github/secrets
```
choover-broad, line 24: Wondering if this should be null? Otherwise the `required` will never print out an error message. I do see that this differs from the Azure pattern, maybe @mflinn-broad can offer some thoughts!

choover-broad (approved):
Sorry, meant to approve this earlier! Left a few inline comments but LGTM

charts/dsp-atlantis/templates/githubCredentials.yaml
```diff
@@ -0,0 +1,25 @@
...
+  # name: name of the secret in KUBERNETES
+  # keysMap: map of keys in the KUBERNETES secret
+  #   some-name-here: name of the key in the secret in KUBERNETES
+  #     path: path of the secret in VAULT
+  #     name: name of the key in the secret in VAULT 
+  name: atlantis-github-providers
+  keysMap:
+    dsp-atlantis-github-token:
+      path: {{ required "A valid .Values.vault.github.pathPrefix is required" .Values.vault.github.pathPrefix }}/tokens
+      key: ATLANTIS-GITHUB-TOKEN
```
choover-broad, line 24: Oh, I just checked and the Vault secret is `ATLANTIS_GITHUB_TOKEN` (underscores)

choover-broad (approved):

charts/dsp-atlantis/values.yaml
jyang-broad: happy to entertain nits, that's how we fix things. I literally just copied the azure line above it without thinking about consequences. `required`, I would prefer to have validation explicitly catch. Maybe commentizing is a good middle ground? 

```
pathPrefix: #/path/to/github/secrets
```

jyang-broad: went again at made the change, not really a good reason to leave as is

charts/dsp-atlantis/templates/githubCredentials.yaml
jyang-broad: fix'd

charts/dsp-atlantis/values.yaml
mflinn-broad: Yeah, sorry about that @jyang-broad. The azure line was me going fast and getting a little sloppy just to get azure stuff working. Didn't mean to trick you into a bad pattern. I'll fix that real quick.

mflinn-broad (approved):
lgtm

PR #404, opened 2021-06-23
DDO-1324 Update Cromiam API version

choover-broad:
It seems the `cloud.google.com/v1beta1` BackendConfig version stopped working in `1.19.10-gke.1600`. Google support has asked us to switch to `cloud.google.com/v1`.

I want to verify it works with cromiam first, afterwards I will update all the BackendConfigs


mflinn-broad (approved):
good thing they gave an obvious warning about this...

PR #405, opened 2021-06-24
DDO-1324 Set API version to `cloud.google.com/v1` in all BackendConfig manifests

choover-broad:
I am probably going to disable the chart autoreleaser before merging and then re-enable it with a commit directly to master... Just to avoid all the noise from failed automerges.


mflinn-broad (approved):
Thanks for dealing with this 😄

PR #407, opened 2021-06-28
Move annotations to be top metadata

knapii-developments:
The linter currently looks for exemptions by looking at annotations associated with the deployment.  Annotations were mistakenly added at the container level which caused builds to break. This PR is a bug fix. 

Succeeding build: https://github.com/broadinstitute/terra-helmfile/pull/1423/checks?check_run_id=2932978845

### DIFF
--- a/charts/workspacemanager/templates/deployment.yaml
+++ b/charts/workspacemanager/templates/deployment.yaml
@@ -5,10 +5,13 @@ metadata:
   name: {{ .Values.name }}-deployment
   labels:
 {{ include "workspacemanager.labels" . | indent 4 }}
-  {{- if and .Values.proxy.enabled .Values.proxy.reloadOnCertUpdate }}
   annotations:
-    reloader.stakater.com/auto: "true"
-  {{- end }}
+    {{- if and .Values.proxy.enabled .Values.proxy.reloadOnCertUpdate }}
+      reloader.stakater.com/auto: "true"
+    {{- end }}
+    {{- range $key, $value := .Values.annotations }}
+    {{ $key }}: {{ $value }}
+    {{- end }}
 spec:
   revisionHistoryLimit: 0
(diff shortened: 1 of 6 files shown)


jack-r-warren (approved):

charts/workspacemanager/templates/deployment.yaml
```diff
@@ -5,10 +5,13 @@ metadata:
   name: {{ .Values.name }}-deployment
   labels:
 {{ include "workspacemanager.labels" . | indent 4 }}
-  {{- if and .Values.proxy.enabled .Values.proxy.reloadOnCertUpdate }}
   annotations:
-    reloader.stakater.com/auto: "true"
-  {{- end }}
+    {{- if and .Values.proxy.enabled .Values.proxy.reloadOnCertUpdate }}
+      reloader.stakater.com/auto: "true"
```
knapii-developments, line 10: @choover-broad @mflinn-broad I changed this logic slightly. However, I don't think it's currently working. I don't see anywhere in Argo where this is true in [workspacemanger-deployment](https://ap-argocd.dsp-devops.broadinstitute.org/applications/workspacemanager-alpha?node=apps%2FDeployment%2Fterra-alpha%2Fworkspacemanager-deployment%2F0) even though based on the [chart](https://github.com/broadinstitute/terra-helm/blob/c26e2b21d1bfc7d84aa074acf1d065c0548d92f8/charts/workspacemanager/values.yaml#L147) it seems like it should be set. Is this expected behavior?

choover-broad: It's disabled in live environments (like `alpha`) but I think it's enabled in personal environments: https://github.com/broadinstitute/terra-helmfile/blob/master/terra/values/workspacemanager/live.yaml#L16

PR #414, opened 2021-07-09
[DDO-1354] Adding liquibase-migration chart to leonardo's

jack-r-warren:
Adds the [liquibase-migration](https://github.com/broadinstitute/terra-helm/tree/master/charts/liquibase-migration) library chart to Leonardo's. Also re-ran the `helm-docs` tool since it seems like no one had run it for a bit.

This won't work--`migration.enabled` shouldn't be set to true anywhere--until https://github.com/broadinstitute/firecloud-develop/pull/2598 has been merged and is in dev. The liquibase-migration chart will expect the liquibase.properties file to already be present in the set of rendered app ctmpl files. 

Once that PR is merged, we can make a terra-helmfile PR with [this diff](https://github.com/broadinstitute/terra-helmfile/compare/DDO-1354-leonardo-liquibase-sync-job) to enable it in dry-run mode for dev (dry-run is defaulted to [here](https://github.com/broadinstitute/terra-helm/pull/414/files#diff-7112dda8e5cc9bfb6e7b7c113a2534b120c2505a761f958bf0ecb14d5a88fff0R131)). We'd merge this PR next, and CI will go and update the chart version in terra-helm so we can drop https://github.com/broadinstitute/terra-helmfile/commit/b84385cda5d3eb0001be4127d893a0fd00f4d0a4 from that helmfile PR (it is there so we can test).

### DIFF
--- a/charts/leonardo/Chart.yaml
+++ b/charts/leonardo/Chart.yaml
@@ -11,3 +11,8 @@ sources:
   - https://github.com/broadinstitute/terra-helm/
   - https://github.com/DataBiosphere/leonardo
   - https://github.com/broadinstitute/leonardo-cron-jobs
+dependencies:
+  - name: liquibase-migration
+    condition: migration.enabled
+    version: 0.2.0
+    repository: https://broadinstitute.github.io/terra-helm/
--- a/charts/leonardo/values.yaml
+++ b/charts/leonardo/values.yaml
@@ -125,7 +125,7 @@ vault:
     dbPasswordKey: "db_password"
 
 migration:
-  # migration.enabled -- (bool) Whether to run a Liquibase migration during sync
+  # migration.enabled -- (bool) Whether to run a Liquibase migration during sync (when enabled, creates K8s resources)
   enabled: false
   # migration.liquibaseCommand -- (string) Liquibase CLI command, can be "updateSQL" for a no-op dry-run, "update" for normal operation
   # See https://docs.liquibase.com/commands/home.html
--- a/charts/leonardo/values.yaml
+++ b/charts/leonardo/values.yaml
@@ -115,3 +115,23 @@ zombieMonitorCron:
 vault:
   # vault.pathPrefix -- (string) Vault path prefix for secrets. Required if vault.enabled.
   pathPrefix:
+  # Migration credentials only referenced if migration.enabled == true
+  migration:
+    # vault.migration.path -- Vault path to secret containing DB credentials
+    path: null
+    # vault.migration.dbUsernameKey -- Key in Vault secret to DB username
+    dbUsernameKey: "db_user"
+    # vault.migration.DbPasswordKey -- Key in Vault secret to DB password
+    dbPasswordKey: "db_password"
+
+migration:
+  # migration.enabled -- (bool) Whether to run a Liquibase migration during sync
+  enabled: false
+  # migration.liquibaseCommand -- (string) Liquibase CLI command, can be "updateSQL" for a no-op dry-run
+  liquibaseCommand: "updateSQL"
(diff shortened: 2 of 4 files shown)


mflinn-broad (approved):

charts/leonardo/values.yaml
```diff
@@ -115,3 +115,23 @@ zombieMonitorCron:
...
+    # vault.migration.path -- Vault path to secret containing DB credentials
+    path: null
+    # vault.migration.dbUsernameKey -- Key in Vault secret to DB username
+    dbUsernameKey: "db_user"
+    # vault.migration.DbPasswordKey -- Key in Vault secret to DB password
+    dbPasswordKey: "db_password"
+
+migration:
+  # migration.enabled -- (bool) Whether to run a Liquibase migration during sync
+  enabled: false
```
Qi77Qi, line 129: is this necessary now that `liquibaseCommand` also governs whether the migration will be run?

charts/leonardo/values.yaml
```diff
@@ -115,3 +115,23 @@ zombieMonitorCron:
...
+  # migration.liquibaseCommand -- (string) Liquibase CLI command, can be "updateSQL" for a no-op dry-run
+  liquibaseCommand: "updateSQL"
```
Qi77Qi, line 131: can you add a link to the doc what possible values are for this?

charts/leonardo/Chart.yaml
```diff
@@ -11,3 +11,8 @@ sources:
   - https://github.com/broadinstitute/terra-helm/
   - https://github.com/DataBiosphere/leonardo
   - https://github.com/broadinstitute/leonardo-cron-jobs
+dependencies:
+  - name: liquibase-migration
+    condition: migration.enabled
```
Qi77Qi, line 16: I don't know helm well, but is it possible to have this condition be something like `liquibaseCommand == updateSQL`?

charts/leonardo/values.yaml
jack-r-warren: `enabled` controls whether any liquibase will be run at all, while `liquibaseCommand` controls what kind of migration will be run. `enabled` being false is how we can safely add this to the chart, because every environment won't suddenly be trying to create the migration job and secret--just where we explicitly enable it, so dev for now. `liquibaseCommand` is how we can test it--we want the job to be created, but we don't want it to start writing to the database on its first go.

charts/leonardo/values.yaml
jack-r-warren: Done, good idea. For normal usage this field might not be even exposed by default in the application values file--for non-testing purposes, the library chart's default of just always providing `update` (a normal, standard migration) would be sufficient, and as you mention above the `enabled` flag controls the behavior that most folks will want to access: whether liquibase runs or not.

charts/leonardo/values.yaml
Qi77Qi: I'm just thinking it could be surprising to ppl, when we need to enable the migration, we need to both update `enable` and update `liquibaseCommand`...maybe add a comment for that?

charts/leonardo/Chart.yaml
jack-r-warren: Since `enabled` controls whether liquibase k8s resources are created, I think that's the correct flag to import the library chart based on. `liquibaseCommand` is going to be either `update` or `updateSQL` for whether we want real operation or dry-run. Just updated the comment in the values file for that too, might be a bit clearer.

Qi77Qi (approved):
lgtm..I think we can probably just try out actual migration whenever you're ready...

jack-r-warren:
Awesome--will be next week, after Jamboree--we'll let you know.

PR #415, opened 2021-07-14
* DDO-1352 Add 30-second sleep

knapii-developments:
Leonardo and Thurloe have a 30-second sleep to avoid CloudSQL proxy time to startup.


charts/leonardo/templates/deployments/_deployment.tpl
```diff
@@ -91,6 +91,11 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.name
+              command: ["/bin/bash"]
+        args:
+        - '-c'
+        - >- # Sleep 30 seconds to allow CloudSQL proxy time to start up.
+          sleep {{ $settings.startupSleep }}
```
jack-r-warren, line 98: - I think `command` needs to be at the same level as `args`, I don't think it goes within `fieldRef`

- I'm not quite sure that this command will actually start the app when the sleep is done--I think supplying a command completely overrides the app image's command instead of just occurring before it ([I think this is pertinent](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes)). I think that's why the charts for other apps like Thurloe and Rawls actually write out the `java` stuff to start the app from bash

PR #422, opened 2021-07-22
Ingress library chart

knapii-developments:
Encapsulates common ingress logic


ingresslibchart/templates/_ingress.yaml
```diff
@@ -0,0 +1,26 @@
...
+spec:
+  tls:
+  - secretName: {{ .Chart.Name }}-cert
+  backend:
+    serviceName: {{ .Chart.Name }}-service
+
+{{- end -}}
+# {{- define "ingresslibchart.configmap" -}}
+# # {{- include "ingresslibchart.util.merge" (append . "ingresslibchart.configmap.tpl") -}}
+# {{- end -}}
```
jack-r-warren, line 26: Are these meant to be commented? This seems to be the only usage of `ingresslibchart.util.merge`

ingresslibchart/templates/_util.tpl
```diff
@@ -0,0 +1,25 @@
...
+{{- $top := first . -}}
+{{- $overrides := fromYaml (include (index . 1) $top) | default (dict ) -}}
+{{- $tpl := fromYaml (include (index . 2) $top) | default (dict ) -}}
+{{- toYaml (merge $overrides $tpl) -}}
+{{- end -}}
+
+{{/*
+Common labels
+*/}}
+{{- define "helm.labels" -}}
```
jack-r-warren, line 18: One idea is rather than calculating out labels yourself, you might be able to grab the labels from the "calling" chart. 

In the liquibase-migration chart, I was able to able to do this:
```
{{- $appName := .Values.migration.appName | default .Release.Name -}}
{{- $labelRef := .Values.migration.labelRef | default (printf "%s.labels" $appName) -}}

// insert other chart stuffs
  labels: {{- include $labelRef . | nindent 4 }}
```

That works because the services themselves define labels like ["buffer.labels"](https://github.com/broadinstitute/terra-helm/blob/master/charts/buffer/templates/ingress/ingress.yaml#L7), and the release [name](https://github.com/broadinstitute/terra-helmfile/blob/master/helmfile.yaml#L118) is set to the name from [versions.yaml](https://github.com/broadinstitute/terra-helmfile/blob/master/versions.yaml#L8). The usage of `default` and stuff is just so that we can override this if we need to--I have yet to find an inconsistency in the naming but I suppose it could happen.

charts/buffer/templates/ingress/ingress.yaml
```diff
@@ -1,23 +1 @@
-{{ if .Values.ingress.enabled -}}
```
jack-r-warren, line 1: Are there any contexts where ingress is disabled? Making a library chart respect an enabled flag is a bit weird, but it is doable I think. Here's how I got it working if you think it is worth it:
- Keep the [if statements](https://github.com/broadinstitute/terra-helm/blob/master/charts/leonardo/templates/migration.yaml#L1) around where you render the template out into the main chart
- Also include the [if statements](https://github.com/broadinstitute/terra-helm/blob/master/charts/liquibase-migration/templates/_job.yaml#L2) in the library chart itself for good measure
- Make the entire dependency [conditional](https://github.com/broadinstitute/terra-helm/blob/master/charts/leonardo/Chart.yaml#L16) on it being enabled

ingresslibchart/templates/_ingress.yaml
```diff
@@ -0,0 +1,26 @@
...
+    ingress.gcp.kubernetes.io/pre-shared-cert: {{ .Values.ingress.cert.preSharedCerts | join "," | quote }}
+{{- end }}
+    {{- range $key, $value := .Values.ingress.annotations }}
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
+spec:
+  tls:
+  - secretName: {{ .Chart.Name }}-cert
+  backend:
+    serviceName: {{ .Chart.Name }}-service
```
jack-r-warren, line 21: Does the servicePort from the original file need to be provided here or is it okay to omit it?

jack-r-warren (commented):
This looks great--left a few ideas from the library chart I played with a little while back just in case it might spare you some experimentation if you want to adopt them

ingresslibchart/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: ingresslibchart
+description: A Helm chart for Ingress
+type: library
+version: 0.1.0
+appVersion: "1.16.0"
```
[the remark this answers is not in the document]
knapii-developments: Nothing. Accidentally carried over from the tutorial I did.

jack-r-warren (approved):

ingresslibchart/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
```
jack-r-warren, line 1: Was looking over this agin in conjunction with #456 and just wanted to note that these files here might be under the wrong directory, since they aren't under the `charts` directory

PR #424, opened 2021-07-28
[DDO-1384] Add dsp-gha-runner chart

jack-r-warren:
Mostly does what it says at https://github.com/actions-runner-controller/actions-runner-controller.

Additions here are:
- pulls secrets from Vault, uses secrets-manager to create the secret that actions-runner-controller expects
- loops over a list to make runners for different repos


mflinn-broad (approved):
LGTM! I didn't check but if the actions-runner-controller chart doesn't accept flags to auto-create the RBAC resources (service account, role... etc) and PSP you may need to create them here also.

jack-r-warren:
Superseded by #431

PR #425, opened 2021-07-28
Support PPW with Rawls - Buffer integration

mflinn-broad:
The purpose of this change is to mount a service account that enables mounting a buffer client SA to the rawls application containers in anticipation of the project per workspace code release on 9/14.

This is essentially a no-op change that mounts a new service account that the project per workspace code base in rawls will reference when it goes out next week.


knapii-developments (approved):

jack-r-warren (approved):
LGTM

PR #427, opened 2021-07-30
[tech debt] bump cert-manager version

jack-r-warren:
Bumps cert-manager version


mflinn-broad (approved):
thanks for fixing my tech debt 😄

PR #429, opened 2021-08-03
DDO-1409 Deploy ArgoCD Notifications alongside ArgoCD

choover-broad:
* Update the dsp-argocd Helm chart to also deploy [ArgoCD Notifications](https://github.com/argoproj-labs/argocd-notifications) (it's necessary to deploy it in the same namespace where ArgoCD creates Application resources)
  * I've configured it in a local subchart instead of an external subchart to save pointless version bump PRs
* Add annotations to the terra-app-generator app to send a Slack notification to #ap-k8s-monitor when a sync fails


mflinn-broad (approved):
nice!

jack-r-warren (approved):
LGTM

PR #432, opened 2021-08-05
[DDO-1158] Add gha-runner charts to Argo

jack-r-warren:
I rendered these out to yaml and applied them manually before merging the charts, adjusting the version numbers here, and making this PR.

- I think slack notifs on unknown sync state--common for a moment if you're using branches and stuff--would be too noisy, so only alerting on straight failures
- defining project separately so as to not tightly couple dsp-ci (the whole cluster) to gha-runners (since I bet someone is going to ask us to run jenkins nodes here within 3 months)
- self heal is on because that seems neat

### DIFF
--- a/charts/dsp-argocd/values.yaml
+++ b/charts/dsp-argocd/values.yaml
@@ -99,3 +99,20 @@ argo-cd:
 # Path in Vault where ArgoCD secrets live
 vault:
   pathPrefix: secret/suitable/argocd
+
+## DRY for projects/apps
+
+projects:
+  dspCi:
+    cluster: https://104.197.236.29
+
+apps:
+  ghaRunner:
+    notiticationChannel: ap-k8s-monitor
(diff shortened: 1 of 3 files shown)


jack-r-warren:
Before helm deploying this I'll delete the applications / project from my manually applied yaml and make sure the CRDs are gone.

charts/dsp-argocd/values.yaml
```diff
@@ -99,3 +99,20 @@ argo-cd:
...
+
+## DRY for projects/apps
+
+projects:
+  dspCi:
+    cluster: https://104.197.236.29
+
+apps:
+  ghaRunner:
+    notiticationChannel: ap-k8s-monitor
```
choover-broad, line 111: nit: typo

choover-broad (approved):
LGTM!

PR #435, opened 2021-08-09
[DDO-1424] Put sysbox on dsp-ci with argocd

jack-r-warren:
Adds default as a dest on the dsp-ci project, puts sysbox there. Also moves the source directory for the existing ones to clean up.

### DIFF
--- a/charts/dsp-argocd/templates/projects/dsp-ci.yaml
+++ b/charts/dsp-argocd/templates/projects/dsp-ci.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   description: CI runners in the broad-dsp-ci cluster; this project managed via dsp-argocd Helm chart
   destinations:
-  - namespace: default
+  - namespace: kube-system
     server: {{ .Values.projects.dspCi.cluster }}
   - namespace: {{ .Values.apps.ghaRunner.namespace }}
     server: {{ .Values.projects.dspCi.cluster }}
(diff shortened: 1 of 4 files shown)


charts/dsp-argocd/templates/projects/dsp-ci.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
 spec:
   description: CI runners in the broad-dsp-ci cluster; this project managed via dsp-argocd Helm chart
   destinations:
+  - namespace: default
```
jack-r-warren, line 8: I'm not grabbing this from `{{ .Values.apps.sysbox.namespace }}` because it is default and that feels dumb, default shouldn't be coupled to sysbox.

mflinn-broad: does sysbox explicitly need to run in default? If so then so be it. If not then slight preference for putting it in it's own namespace.

mflinn-broad (approved):
LGTM! For the record this is a pretty low risk argo change from what I can tell. It is only changing the crds that argo reads, not the argo deployment itself.

charts/dsp-argocd/templates/projects/dsp-ci.yaml
jack-r-warren: It actually puts everything in kube-system explicitly. I don't think any actual resources will get created in whatever namespace we set here. Should I actually set  kube-system here or try changing their manifests to respect the deployment namespace?

mflinn-broad: In that case you may need to grant the argo project explicit permissions to install to kube-system. I *think* it can only apply resources in namespaces explicitly listed under destinations.

jack-r-warren: Oh then that's a very good point, I'll change the dest. I think their manifests do this as partly a "you can really really only ever have exactly one of these installed at once" deal

PR #437, opened 2021-08-10
[DDO-1424] name sysbox psp correctly, allow host paths

jack-r-warren:
Hopefully this will be the only one of these. Helps to name the psp the same as what the role actually references, and of course hostPaths are their own field here.


choover-broad (approved):
👍

PR #440, opened 2021-08-10
[DDO-1425] Improve runner instances chart

jack-r-warren:
Big changes to values file here because there's a required annotation that the sysbox stuff needs. So, here's what we have now:

- `.Values.runnerDefaults` has metadata and spec keys, put whatever in there, example:

```yaml
runnerDefaults:
  metadata:
    annotations:
      my-annotation: foo
  spec:
    imagePullPolicy: Always
```

- There's a `.Values.defaultDockerConfig` string field (override per runner via .dockerConfig). That is used as a key for the `.Values.dockerConfigs` object, where each key has metadata and spec keys, example:

```yaml
dockerConfigs:
  privilegedDockerInDocker:
    metadata:
    spec:
      dockerEnabled: false
      dockerdWithinRunnerContainer: true

defaultDockerConfig: privilegedDockerInDocker
```

- Finally, there is a `configOverrides` field that can be provided per-runner with, you guessed it, metadata and spec keys, example:

```yaml
runners:
  - name: example
    repository: broadinstitute/example-repo
    configOverrides:
      metadata:
        annotations:
          my-annotation: foo
      spec:
        imagePullPolicy: Always
```

- At the end, all three objects are deep merged together and the values for metadata and spec are pulled out and used if they're not empty. With the exception of the dockerConfig / defaultDockerConfig thing (already set in the base file here) it is all optional.

The reason for all of this is that the whole thing now gets configured via the values file but it isn't super verbose either. We can do anything from "one-line change to make every runner suddenly use a privileged sidecar for docker" to "tweak the uid size for a particular runner's usage of cri-o/sysbox" to "use some successor to sysbox years from now"


mflinn-broad (approved):
lgtm!

PR #442, opened 2021-08-13
DDO-1430 Investigate reducing metric ingest spend

mflinn-broad:
This pr reduces the frequency with which prometheus writes new data points to timeseries from every 30 seconds to every 60 seconds. This is one of a few different measures we can take to try to reduce our metric ingest spend in GCP


charts/terra-prometheus/README.md
```diff
@@ -20,44 +20,6 @@ a helm chart to deploy monitoring infrastructure
...
-    text: >-
-      {{ range .Alerts }}
-        *Alert:* `{{ .Labels.severity }}`
-        *Description:* {{ .Annotations.message }}
-        *Graph:* <{{ .GeneratorURL }}|:chart_with_upwards_trend:> *Runbook:* <{{ .Annotations.runbook }}|:spiral_note_pad:>
-        *Details:*
-        {{ range .Labels.SortedPairs }} • *{{ .Name }}:* `{{ .Value }}`
-        {{ end }}
-      {{ end }}
-```
```
jack-r-warren, line 59: Are we good to remove these docs?

jack-r-warren (approved):

charts/terra-prometheus/README.md
mflinn-broad: nope. just noticed this, fixing

jack-r-warren (approved):

PR #445, opened 2021-08-13
[DDO-1425] Bump gha-runner-instances to use our image

jack-r-warren:
Tested that it works by having Helm use 0.3.0 manually. Even if we need to change the image in the future, this pulls nightly of it, and we can override it in the values file in terra-helm-definitions.


jyang-broad (approved):
2 + 1 = 3, math checks out

PR #450, opened 2021-08-17
[DDO-1449] Add wsmtest-sync account

jack-r-warren:
Adds an account to Argo CD able to sync wsmtest. I've generated a token for it and fully tested that this works, and that it rejects attempts to do anything else.

```
curl --fail --silent --show-error --location --request POST 'https://ap-argocd.dsp-devops.broadinstitute.org/api/v1/applications/workspacemanager-wsmtest/sync' \
    --header "Authorization: Bearer $(vault read -field=token secret/suitable/argocd/local-accounts/low-privileged/wsmtest-sync)" \
    | jq .operation
```

### DIFF
--- a/charts/dsp-argocd/values.yaml
+++ b/charts/dsp-argocd/values.yaml
@@ -23,6 +23,7 @@ argo-cd:
       accounts.fcprod-jenkins: apiKey
       accounts.local-notsuitable: login
       accounts.local-suitable: login
+      accounts.wsmtest-sync: apiKey
       configManagementPlugins: |
         - name: legacy-configs
           generate:
@@ -66,6 +67,8 @@ argo-cd:
         p, role:terra-ci, applications, sync, */*, allow
         p, role:terra-ci, applications, action/apps/Deployment/restart, */*, allow
         g, terra-ci, role:terra-ci
+        p, role:wsmtest-sync, applications, sync, terra-wsmtest/workspacemanager-wsmtest, allow
+        g, wsmtest-sync, role:wsmtest-sync
   controller:
     extraArgs:
     - --repo-server-timeout-seconds=180


em-may (approved):
Looks simple enough, would appreciate a brief explanation of what's going on so I can better understand, thanks!

jack-r-warren:
> Looks simple enough, would appreciate a brief explanation of what's going on so I can better understand, thanks!

Well, I don't know how fully I understand either. In theory, I'm just following [this](https://argoproj.github.io/argo-cd/operator-manual/user-management/#create-new-user) and [this](https://argo-cd.readthedocs.io/en/stable/operator-manual/rbac/). That seems simple enough except neither Mike or I are positive how non-devops people who log in with github get access, because there's not a github-team-to-argo-role binding for non-devops people in that policy.csv snippet we have.

We decided it is a good question but not really relevant for this so it can wait for tomorrow.

choover-broad (approved):
Looks fine!

An alternative option would be to set this permission directly in the wsmtest project ([example](https://github.com/broadinstitute/terra-helm/blob/master/charts/terra-argocd-env/templates/project.yaml#L17)): ) but that would involve some helmfile magic ([hereabouts](https://github.com/broadinstitute/terra-helmfile/blob/master/helmfile.yaml#L176)) and this is a one-off case, so this is probably better!

jack-r-warren:
> Looks fine!
> 
> An alternative option would be to set this permission directly in the wsmtest project ([example](https://github.com/broadinstitute/terra-helm/blob/master/charts/terra-argocd-env/templates/project.yaml#L17)) but that would involve some helmfile magic ([hereabouts](https://github.com/broadinstitute/terra-helmfile/blob/master/helmfile.yaml#L176)) and this is a one-off case, so your approach here is probably better!

Mm that's good to know, if we get a bunch of requests for this sorta thing maybe that's the way we build this out

PR #451, opened 2021-08-17
POL-51 ECM background processes

dvoet:
https://broadworkbench.atlassian.net/browse/POL-51

deployment changes for frontend and backend ECM

### DIFF
--- a/charts/externalcreds/templates/deployment.yaml
+++ b/charts/externalcreds/templates/deployment.yaml
@@ -183,10 +183,10 @@ spec:
           {{- toYaml $.Values.probes.startup.spec | nindent 10 }}
         {{- end }}
       {{- if not $.Values.postgres.enabled }}
-        command: |-
-            {{- range .command }}
-          - {{ . | quote }}
-            {{- end }}
+        command:
+          {{- range .command }}
+        - {{ . | quote }}
+          {{- end }}
       - name: cloudsql-proxy
         image: gcr.io/cloudsql-docker/gce-proxy:1.16
         env:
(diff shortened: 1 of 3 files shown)


charts/externalcreds/templates/deployment.yaml
```diff
@@ -1,11 +1,13 @@
-{{- $imageTag := .Values.imageConfig.tag | default .Values.global.applicationVersion -}}
+{{- $imageTag := $.Values.imageConfig.tag | default $.Values.global.applicationVersion -}}
+{{- range $.Values.appDeployments }}
```
dvoet, line 2: This `range` function will create the included block for each `appDeployments` in values.yaml. Most of the changes below are changing `.` (current scope) to `$` (root scope) because `range` creates a new scope so the current scope is no longer the root scope as it was before.

charts/externalcreds/templates/deployment.yaml
```diff
@@ -161,43 +163,44 @@ spec:
...
         {{- end }}
-        {{- if .Values.probes.startup.enabled }}
+        {{- if $.Values.probes.startup.enabled }}
         startupProbe:
-          {{- toYaml .Values.probes.startup.spec | nindent 10 }}
+          {{- toYaml $.Values.probes.startup.spec | nindent 10 }}
         {{- end }}
-      {{- if not .Values.postgres.enabled }}
+      {{- if not $.Values.postgres.enabled }}
+        command: [{{ join "," .command }}]
```
dvoet, line 186: command was introduced here because frontend and backend differ

charts/externalcreds/templates/ingress/service.yaml
```diff
@@ -15,7 +15,7 @@ metadata:
 spec:
   type: ClusterIP
   selector:
-    deployment: {{ .Values.name }}-deployment
+    deployment: {{ .Values.name }}-frontend-deployment
```
dvoet, line 18: select only frontend for ingress

PR #452, opened 2021-08-19
Add ingress lib

knapii-developments:
Ingress library.


charts/ingresslibchart/templates/_ingress.yaml
```diff
@@ -0,0 +1,23 @@
+{{- define "libchart.ingress.tpl" -}}
+apiVersion: networking.k8s.io/v1beta1
```
jack-r-warren, line 2: Is there a nonbeta version we can use? I've gotten some warnings here and there about beta ingress being deprecated and unavailable in future k8s versions.

charts/ingresslibchart/templates/_ingress.yaml
```diff
@@ -0,0 +1,23 @@
...
+kind: Ingress
+metadata:
+  name: {{ .Chart.Name }}-ingress
+  labels:
+    {{- include "helm.labels" . | nindent 4 }}
+  annotations:
+    networking.gke.io/v1beta1.FrontendConfig: "{{ .Chart.Name }}-ingress-frontendconfig"
+    kubernetes.io/ingress.global-static-ip-name: {{ required "ingress.staticIpName value is required" .Values.ingress.staticIpName | quote }}
```
jack-r-warren, line 10: I don't know if it makes sense here, but documenting the expected values file structure might be helpful. I know a values file won't be included as defaults or anything because it's a library chart but maybe even an example in a readme or something would help

charts/ingresslibchart/templates/_ingress.yaml
```diff
@@ -0,0 +1,23 @@
...
+{{- end }}
+    kubernetes.io/ingress.allow-http: "false"
+spec:
+  backend:
+    serviceName: {{ .Chart.Name }}-service
+    servicePort: 443
+{{- end -}}
+
+{{- define "libchart.ingress" -}}
+{{- include "libchart.util.merge" (append . "libchart.ingress.tpl") -}}
```
jack-r-warren, line 22: I think my helm syntax might be rusty, what is being merged into what here?

jack-r-warren (approved):

charts/ingresslibchart/templates/_ingress.yaml
knapii-developments: This merges values that are not included in the library. Here's an [example](https://github.com/broadinstitute/terra-helm/blob/a3fd519477ba6594a5d86acfe0b7e67cf2022282/charts/externalcreds/templates/ingress/ingress.yaml#L2). Here, externalcred as additional values that are not in the library.

charts/ingresslibchart/templates/_ingress.yaml
knapii-developments: Right now, we have no value files structure. However, when looking at how other applications define ingress values, we seem to be aligned. The error message "ingress.staticIpName" says where the value should be located.

charts/ingresslibchart/templates/_ingress.yaml
```diff
@@ -0,0 +1,23 @@
+{{- define "libchart.ingress.tpl" -}}
+apiVersion: networking.k8s.io/v1
```
knapii-developments, line 2: @jack-r-warren Updated the API version

PR #454, opened 2021-08-20
[DDO-1347] Revere chart

jack-r-warren:
Took me forever but this _seems_ vaguely correct. I'm mainly interested in what here we should take out--I incorporated the ability to enable a lot of [Denis's suggestions](https://broadworkbench.atlassian.net/browse/DDO-1393?focusedCommentId=49682), but there's some extra maintenance there.

The only thing missing here is a job to run the `prepare` step. I'm specifically holding off on writing that because I will need to debug the NetworkPolicy stuff first, and I want the syncs to just complete normally in the meantime. It will be almost identical to the deployment spec, to the point that I'll probably try to abstract it.


charts/revere/values.yaml
```diff
@@ -0,0 +1,211 @@
...
+    # -- (string) Memory to request for the deployment
+    memory: 4Gi
+  limits:
+    # -- (string) Number of CPU units to limit the deployment to
+    cpu: 500m
+    # -- (string) Memory to limit the deployment to
+    memory: 4Gi
+
+probes:
+  readiness:
```
knapii-developments, line 43: Do you plan on adding any configurations to these probes? I think these should mirror how [we define probes in other charts](https://github.com/broadinstitute/terra-helm/blob/d1ee696637052dc2c35ef3b06e21f2974ec7e935/charts/buffer/values.yaml#L39)s.

charts/revere/templates/deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+        {{- /* Automatically restart deployments on config map change: */}}
+        checksum/{{ .Values.name }}-cm: {{ include (print $.Template.BasePath "/configMap.yaml") . | sha256sum }}
+        {{- range $key, $value := .Values.annotations }}
+        {{ $key }}: {{ $value }}
+        {{- end }}
+    spec:
+      serviceAccountName: {{ .Values.name }}-sa
+
+      {{- if .Values.runtimeClassName }}
+      runtimeClassName: {{ .Values.runtimeClassName }}
```
mflinn-broad, line 33: Is this related to one of Dennis' suggestion. I don't recall seeing this field before?

charts/revere/values.yaml
```diff
@@ -0,0 +1,211 @@
...
+    enabled: false
+    # -- (object) Spec for the liveness probe
+    spec:
+  startup:
+    # -- (boolean) If the startup probe should be enabled
+    enabled: false
+    # -- (object) Spec for the startup probe
+    spec:
+
+secrets:
```
knapii-developments, line 59: I don't think you're supposed to put secrets in the value file. Normally you'd store these as environment variables or use helm secrets. You can read more about them [here](https://learning.oreilly.com/library/view/learn-helm/9781839214295/B15458_09_Final_JM_ePub.xhtml#_idParaDest-203)

charts/revere/templates/deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+        {{- if .Values.probes.readiness.enabled }}
+        readinessProbe:
+          {{- toYaml .Values.probes.readiness.spec | nindent 10 }}
+        {{- end }}
+        {{- if .Values.probes.liveness.enabled }}
+        livenessProbe:
+          {{- toYaml .Values.probes.liveness.spec | nindent 10 }}
+        {{- end }}
+        {{- if .Values.probes.startup.enabled }}
+        startupProbe:
```
mflinn-broad, line 84: Not sure if you need this sense we don't need to worry about interrupted liquibase migrations here :)

charts/revere/templates/networkPolicy.yaml
```diff
@@ -0,0 +1,69 @@
...
+metadata:
+  name: {{ .Release.name }}-networkpolicy
+  labels:
+    {{- include "revere.labels" . | nindent 4 }}
+spec:
+  policyTypes:
+  {{- if .Values.networkPolicy.ingress.enabled }}
+  - Ingress
+  {{- end }}
+  {{- if .Values.networkPolicy.egress.enabled }}
```
mflinn-broad, line 12: Does App sec care about both ingress and egress here?

charts/revere/templates/secretDefinition.yaml
```diff
@@ -0,0 +1,38 @@
+{{- if .Values.secrets.gcpServiceAccount.secretsManager.enabled }}
+---
+apiVersion: secrets-manager.tuenti.io/v1alpha1
+kind: SecretDefinition
```
knapii-developments, line 4: Can you provide a link to this kind of manifest?

charts/revere/templates/networkPolicy.yaml
```diff
@@ -0,0 +1,69 @@
...
+      {{- if .Values.secrets.gcpServiceAccount.workloadIdentity.enabled }}
+      # GKE Metadata Server
+      - port: 988
+        protocol: TCP
+      {{- end }}
+      # DNS
+      - port: 53
+        protocol: TCP
+      - port: 53
+        protocol: UDP
```
mflinn-broad, line 68: This is an explicit allow rather than explicit deny correct?

charts/revere/values.yaml
jack-r-warren: At some point, definitely! Revere doesn't have any ingress at all right now--it literally doesn't run a server at all, it is just a client of Statuspage.io and Google PubSub. Once it runs a server or something we'll have to come through and add that capability to the chart anyway, I guess I just wanted these placeholders here so I could add probes for testing purposes.

charts/revere/templates/workloadIdentity.yaml
```diff
@@ -0,0 +1,27 @@
...
+    {{- include "revere.labels" . | nindent 4 }}
+---
+apiVersion: iam.cnrm.cloud.google.com/v1beta1
+kind: IAMPolicyMember
+metadata:
+  name: {{ .Values.name }}-iampolicy-member
+  labels:
+    {{- include "revere.labels" . | nindent 4 }}
+spec:
+  member: serviceAccount:{{ .Values.secrets.gcpServiceAccount.workloadIdentity.projectId }}.svc.id.goog[{{ .Release.Namespace }}/{{ .Values.name }}-sa]
```
mflinn-broad, line 20: How does the namespace get used here? If this is necessary so be it, but it's good to be careful when using `.Release.Namespace` is it can unintentional bake in assumptions about where an application is deployed.

charts/revere/values.yaml
```diff
@@ -0,0 +1,211 @@
...
+    secretsManager:
+      # -- (boolean) If secretsManager should be used for the GCP SA
+      enabled: false
+      # -- (string) Path within Vault where the SA's key is stored
+      vaultPath:
+      # -- (string) Field name within the secret for the SA's key
+      vaultKey:
+      # -- (string) Encoding of the Vault field (either `text` or `base64`)
+      sourceEncoding: base64
+  statuspageApiKey:
```
mflinn-broad, line 77: how do we get this if secrets manager is not enabled?

charts/revere/values.yaml
```diff
@@ -0,0 +1,211 @@
...
+    cidrAllowList:
+    # -- (list) Posts/protocols to permit for ingress
+    portAllowList:
+  egress:
+    # -- (boolean) If the NetworkPolicy should be enabled for egress
+    enabled: true
+    # -- (list) CIDR ranges to permit for egress
+    cidrAllowList:
+    # Atlassian for Statuspage, per https://ip-ranges.atlassian.com/
+    - 2600:1f18:2146:e300::/56
```
mflinn-broad, line 102: Not sure how I feel about encoding all of this as default values. I think for the helm chart I'd prefer just an empty list as default. Then we populate this in terra-helmfile.

mflinn-broad (approved):
Meant to approve. This all looks really solid to me from a helm perspective. Most of my questions are just due to my lack of familiarity with some of the new resource types being used here

charts/revere/values.yaml
jack-r-warren: Of course! None of the fields here are meant to contain secrets themselves, they're just mechanisms for porting secrets into the deployment.

Workload Identity is a way for Google to allow Kubernetes service accounts to be linked to Google service accounts. Denis suggested we use it because it is more secure than storing an SA's private key in Vault. We might still use Vault for it, though, so that's why secrets-manager that we use everywhere else is also an option. I'm not calling it `vault` because I'm trying to be a bit forward-looking (other Vault mechanisms like the argocd-vault-plugin might require us to pass things differently).

In other words, the fields here are like `.Values.secrets.${The thing that is secret}.${How that thing gets into the deployment}`, so `.Values.secrets.gcpServiceAccount.workloadIdentity` configures Workload Identity to connect the GCP SA to the deployment. I just updated some of the comments here to make it a bit more clear.

I was admittedly having some trouble naming fields here, I'm open to any suggestions you have.

charts/revere/templates/secretDefinition.yaml
jack-r-warren: Totally, this comes from [secrets-manager](https://github.com/tuenti/secrets-manager), we use it whenever we don't get secrets out of firecloud-develop and get it out of Vault directly. [Atlantis](https://github.com/broadinstitute/terra-helm/blob/master/charts/dsp-atlantis/templates/cleanup/secretdefinition.yaml), [duos](https://github.com/broadinstitute/terra-helm/blob/master/charts/duos/templates/secrets.yaml), [buffer](https://github.com/broadinstitute/terra-helm/blob/master/charts/buffer/templates/secrets.yaml), and [wsm](https://github.com/broadinstitute/terra-helm/blob/master/charts/workspacemanager/templates/secrets.yaml) all use it, I'm sure there's other cases.

charts/revere/templates/workloadIdentity.yaml
mflinn-broad: I realize it shows up in some other helm charts too 😄 That is a small tech debt issue to address.

charts/revere/templates/deployment.yaml
jack-r-warren: Yeah this is GKE sandbox capability. From an app's perspective, you just specify the runtime class ([link](https://cloud.google.com/kubernetes-engine/docs/how-to/sandbox-pods#sandboxed-application)), but it is more work to set up on a cluster (it is a flag you enable on a new node pool). I figure it would be a pretty minimal thing to expose here and we can discuss if adding a node pool is a route we want to go down.

charts/revere/templates/deployment.yaml
jack-r-warren: I didn't even realize that, that's very true. I removed the startup one, are both of the other ones still useful? Have never worked with probes before.

charts/revere/templates/deployment.yaml
mflinn-broad: cool seems simple enough. Adding node pools is relatively easy via our cluster terraform module.

charts/revere/templates/deployment.yaml
mflinn-broad: yes the others are useful! Readiness will remove a pod from a loadbalancer pool if a health check fails for x amount of time. Liveness will restart the entire pod if a health check is failing for y amount of time. 

Generally the approach we take is the readiness probe will trigger and remove a pod from the LB pool after 2 minutes of being unhealthy. The Liveness probe will kick in after 5 minutes and kill the pod entirely. After 10 minutes we trigger pagerduty.

charts/revere/templates/networkPolicy.yaml
jack-r-warren: Egress is how we can prevent Revere from talking to ArgoCD, so that's the main one they want, even though egress is much harder than ingress to maintain. As for ingress, it is a way we could let only VPNed Broadies access Revere. I don't know if we'd implement that here or with a more classical app armor rule.

charts/revere/templates/networkPolicy.yaml
jack-r-warren: Yeah, the second a network policy for ingress or egress applies to a pod, ingress or egress is denied and then you have to add it back in here. It is separate for ingress and egress, at least.

charts/revere/templates/networkPolicy.yaml
mflinn-broad: This was a poorly phrased question by me. I guess what I'm really asking is can these egress configs prevent intra cluster communications? Ie over kubernetes internal DNS? Also if this question doesn't make sense that is totally fine, we can chat about it next week.

charts/revere/templates/workloadIdentity.yaml
jack-r-warren: It is unfortunately very necessary, [this](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#identity_sameness_across_clusters) helps explain why and [step 7 here](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#authenticating_to) demos it being necessary whether you use gcloud or a manifest to do it.

charts/revere/values.yaml
jack-r-warren: Right now, we don't, it is fairly required. I think I might be future-proofing too much here, I think if we used another Vault-to-k8s mechanism the values might change so that's why I was already preparing to be able to toggle between secrets-manager and some other Vault mechanism

charts/revere/values.yaml
jack-r-warren: That's a really good point--I think I did all the research to flesh it out and didn't want to lose the info here. I can split this out to a terra-helm-definitions PR on Monday

charts/revere/values.yaml
mflinn-broad: > I think if we used another Vault-to-k8s mechanism

I think that's highly likely down the road. So I'm totally ok with this. Just confirming there wasn't another method of getting the secret I was missing.

charts/revere/values.yaml
mflinn-broad: Also we may want to take advantage of Chelsea's recent helmfile changes and deploy this via terra-helmfile rather than terra-helm-definitions.

charts/revere/templates/networkPolicy.yaml
jack-r-warren: Maybe this might be a question to ask Denis. The k8s docs and every tutorial out there talk about different mechanisms for allowing pods to talk to each other through network policies, implying that it does indeed block in-cluster traffic, but it is a good point that we actually care about the opposite, we want a bulletproof firewall

charts/revere/values.yaml
jack-r-warren: Good idea. I have a -definitions PR open [here](https://github.com/broadinstitute/terra-helm-definitions/pull/67) just as an example but I'm marking it as a draft--I think I might tackle that once the chart's design is a bit more fleshed out but before merging anything, just so we can sorta be reviewing one thing at a time.

knapii-developments (approved):

jack-r-warren:
Merging this to avoid it becoming stale between now and when I come back, but explicitly _not_ merging https://github.com/broadinstitute/terra-helm-definitions/pull/67 because when we eventually do deploy this we'll do so via terra-helmfile (https://github.com/broadinstitute/terra-helm/pull/454#discussion_r694138147 for context)

PR #456, opened 2021-08-23
* DDO-993 Add front-, back- end configs and service library charts

knapii-developments:


charts/ingresslibchart/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v2
+name: ingress-library
```
choover-broad: Quick comment on naming consistency.

Can the name be `ingress` or `ingresslib`? And can the chart directory match the name instead of `ingresslibchart`?

charts/ingresslibchart/templates/_backendconfig.yaml
```diff
@@ -0,0 +1,22 @@
+{{- define "libchart.backendconfig.tpl" -}}
```
choover-broad, line 1: Another naming consistency comment.

* Can the template prefix also match the chart name? Eg. `ingresslib.backendconfig` instead of `libchart.backendconfig`?
* Can the file be named `_backedconfig.tpl` instead of `_backedconfig.yaml`? (Since it's a helper)
* Can `.tpl` be removed from the template name (the `define` statement)? Just for the sake of following convention

charts/ingresslibchart/templates/_frontendconfig.yaml
```diff
@@ -0,0 +1,16 @@
+{{- define "libchart.frontendconfig.tpl" -}}
+apiVersion: networking.gke.io/v1beta1
+kind: FrontendConfig
+metadata:
+  name: {{ .Chart.Name  }}-ingress-frontendconfig
+  labels:
+    {{- include "helm.labels" . | nindent 4 }}
```
choover-broad, line 7: We'll want to include the labels of the parent chart here, so maybe it makes sense to pass that in as a parameter? The liquibase migration chart has a good example for how to do this: https://github.com/broadinstitute/terra-helm/blob/master/charts/liquibase-migration/templates/_job.yaml#L7

charts/ingresslibchart/templates/_backendconfig.yaml
```diff
@@ -0,0 +1,22 @@
...
+    timeoutSec: 5
+    healthyThreshold: 2
+    unhealthyThreshold: 2
+    type: HTTPS
+    port: 443
+    requestPath: {{ .Values.ingress.requestPath }}
+{{- end }}
+
+{{- define "libchart.backendconfig" -}}
+{{- include "libchart.util.merge" (append . "libchart.backendconfig.tpl") -}}
```
choover-broad, line 21: hm, what's happening here? a bit confused why the merge is necessary!

choover-broad (commented):
Hey @knapii-developments, good progress here! I left some preliminary comments, will come back for another look later

charts/ingresslibchart/templates/_backendconfig.yaml
knapii-developments: From my understanding .[tpl files are used when a template isn't rendered](https://helm.sh/docs/chart_template_guide/named_templates/). libchart.backendconfig.tpl is just the boilerplate, while [this](https://github.com/broadinstitute/terra-helm/blob/cde92ae3c01746d3a0d56a02d2b048004316763a/charts/ingresslibchart/templates/_backendconfig.yaml#L20) actually defines and outputs the chart. The .tpl extension also helps others who come to the code realize that it's not rending anything.[ Helm library documentation](https://helm.sh/docs/chart_template_guide/named_templates/) also follows this naming convention as well. Same applies for the file extension.

jack-r-warren (approved):
I think this looks generally good. I know this is split across multiple PRs and that's a pain to refactor so I haven't focused in on it much, but I think my note from https://github.com/broadinstitute/terra-helm/pull/422#discussion_r675713160 still stands about naming and redefining labels.

I don't think it is a big deal--we just should be careful in the future to remember that kinda generic things like `helm.labels`, `helm.selectorLabels`, or `libchart.service` are actually specifically defined by the ingress library chart.

This sorta goes with https://github.com/broadinstitute/terra-helm/pull/452#discussion_r692360345 too--I totally get that we don't have an existing standard for any of this stuff; I think that maybe writing down values file expectations or chart definitions in readmes or otherwise can help actually establish those standards.

charts/ingresslibchart/templates/_frontendconfig.yaml
knapii-developments: I noticed that the parent label across charts is repeated and has the same definitions. In general, if additional information needs to be added, they can be as seen [here](https://github.com/broadinstitute/terra-helm/blob/08c549ab09e7904ad37ec8db6d9f89daf25e5164/charts/consent/templates/ingress.yaml#L4). Also, if there are repeated values between the charts, the parent label will take precedence. Hence, there's no need to pass it as a parameter.

jack-r-warren: I think this bit in particular is split across multiple PRs--https://github.com/broadinstitute/terra-helm/pull/422#discussion_r675713160 also notes this

charts/ingresslibchart/templates/_backendconfig.yaml
jack-r-warren: She said it had to do with the values file here (https://github.com/broadinstitute/terra-helm/pull/452#discussion_r692349375), but not totally sure I'm understanding this yet either -- I totally get that the merge is necessary at the call site, I'm just not totally sure about it within the chart

charts/ingresslibchart/templates/_backendconfig.yaml
choover-broad: Hey @knapii-developments , I think this is one reason why I'm confused!

I thought that library charts *only* included named templates, and don't render any manifests directly. I.e., the library chart includes a bunch of files like `_foo.tpl` and then the downstream charts use `include` or `template` to render the named templates from the library.

choover-broad: So I guess I just need some clarification: is this chart intended to render things like BackendConfigs directly? Or should it only include named templates, which are then used in downstream charts like `agora`, with `include` statements?

choover-broad (approved):
Approving because I don't think that merging this PR will cause any issues, especially since I don't think it's actually used anywhere.

I would appreciate a bit of clarification on whether this is technically a library chart (named templates only) or whether it's more of a subchart, but that's not a merge blocker.

(This also might make more sense to me if there was an example of a chart using it)

charts/ingresslibchart/templates/_backendconfig.yaml
knapii-developments: From [Learning Helm](https://learning.oreilly.com/library/view/learning-helm/9781492083641/ch06.html#idm46125990551032) with the files substituted in.

> libchart.backendconfig.tpl contains a template for a resource. This will look similar to other templates. It provides a blueprint that is meant to be overridden by the caller in a chart that includes this library. libchart.backendconfig is a special template. This is the template another chart will use. It takes libchart.backendconfig.tpl along with another template, yet to be defined, containing overrides, and merges them into one output.libchart.backendconfig uses a utility function that handles the merging and is handy to reuse.

This implementation is also used in the Helm library documentation.

charts/ingresslibchart/templates/_backendconfig.yaml
knapii-developments: Let me know if this [comment](https://github.com/broadinstitute/terra-helm/pull/456#discussion_r694172569) makes sense

knapii-developments:
> Approving because I don't think that merging this PR will cause any issues, especially since I don't think it's actually used anywhere.
> 
> I would appreciate a bit of clarification on whether this is technically a library chart (named templates only) or whether it's more of a subchart, but that's not a merge blocker.
> 
> (This also might make more sense to me if there was an example of a chart using it)

It's a library chart because it  _"defines chart primitives or definitions which can be shared by Helm templates in other charts."_  However, the library does consist of sub/named templates.

charts/ingresslibchart/templates/_backendconfig.yaml
knapii-developments: Overall, IDK. I just followed the helm library documentation haha

charts/ingresslibchart/Chart.yaml
knapii-developments: 0914d3a9

charts/ingresslibchart/templates/_backendconfig.yaml
jack-r-warren: I think I understand now… TIL about currying in go templates. So `libchart.util.merge` is a function of three arguments (given as a list), and `libchart.backendconfig` is a function of two arguments (also given as a list)--callers provide ["current context" and "named variable that's either empty or has overrides"](https://github.com/broadinstitute/terra-helm/pull/457/files#diff-59cf366944762a86e9b3df426e5004b7655590b3bd7a1d3ed51fefceeae942d8R1) meanwhile line 21 above provides ["defaults"](https://github.com/broadinstitute/terra-helm/pull/457/files#diff-9183a237943b4944f355c6be7ffd11e7e58b59dc1bb7054dfb0293aec30461f6R6). Sorry for not groking this, I don't have O'Reilly and haven't seen this used in the wild before.

charts/ingresslibchart/templates/_frontendconfig.yaml
jack-r-warren: I think I understand the merge mechanic now, but I think the problem remains if an app uses different or fewer labels than what is defined by `helm.labels`. If we pass by parameter, it works as expected; if this chart also repeats some expectations about labels, then the end labels become a union of the base chart and what this library provides.

I'm totally with you on the labels being frequently repeated, and I think a library chart is a great way to reduce that repetition--I just don't think that an ingress library chart that isn't common across all apps is the right place to do that abstraction. I think we have an opportunity here to _not_ write down another repetition of the labels, and we can do the pass-by-reference trick to make it easier to abstract in the future.

PR #462, opened 2021-08-26
Ddo 1399 leonardo preview chart changes

jyang-broad:
update to the leonardo chart to support preview envs, will not affect current envs the two switches need to be enabled to make changes.

### DIFF
--- a/charts/leonardo/Chart.yaml
+++ b/charts/leonardo/Chart.yaml
@@ -18,5 +18,5 @@ dependencies:
     repository: https://broadinstitute.github.io/terra-helm/
   - name: mysql
     condition: localMysql.enabled
-    version: 0.1.0
+    version: 0.1.0-461.1630012760.fea8cb
     repository: https://broadinstitute.github.io/terra-helm/


charts/leonardo/values.yaml
```diff
@@ -1,6 +1,9 @@
 global:
   # global.applicationVersion -- What version of the Leonardo application to deploy
   applicationVersion: latest
+# Use a k8s instance of mysql for database, used for preview environments.
```
knapii-developments, line 4: Can you move this to be under deployments?

charts/mysql/values.yaml
```diff
@@ -0,0 +1,82 @@
+# Default values for mysql containers, can be overwritten at build time by environment configurations
+
+global:
+  # global.storageClass -- Storage class to use when provisioning persistent disks
+  storageClass: standard
+  # global.applicationVersion -- What version of mysql to deploy
+  applicationVersion: latest
+
+# name -- A name for the statefulset that will be substituted into resource definitions
```
knapii-developments, line 9: I know we don't have a standard when it comes to organizing the values file. However, most of them have deployment values under a deployment key. Can you do the same here?

charts/leonardo/values.yaml
jyang-broad: I was about to do this and then realized that doesn't work for this variable. the localMysql actually sets up a totally different set of containers statefulset/service/etc and doesn't actually change the deployment at all. (The config changes to the deployment actually happen in firecloud-develop configs.)

charts/mysql/values.yaml
jyang-broad: Yes! Thanks for catching this. I'm actually going to move this to a new ticket and then poke you on this b/c I have to update the chart, which would blow up this PR.

PR #467, opened 2021-09-01
* DDO-1352 Add 30-second sleep for Consent couldSQL proxy

knapii-developments:
Env-dif can be seen [here](https://github.com/broadinstitute/terra-helmfile/pull/1621#issuecomment-910542814).

### DIFF
--- a/charts/consent/README.md
+++ b/charts/consent/README.md
@@ -2,7 +2,7 @@
 
 A Helm chart for DUOS Consent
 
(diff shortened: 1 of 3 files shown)


choover-broad (approved):
This looks great @knapii-developments! As a reviewer it's much easier to see what's going on when the change is only one chart haha

Thanks for trying this out, can't wait to see it in action!

PR #468, opened 2021-09-02
* DDO-1352 Add 30-second sleep for CRL Janitor couldSQL proxy

knapii-developments:
Env-dif can be seen [here](https://github.com/broadinstitute/terra-helmfile/pull/1625/checks?check_run_id=3495447673#step:8:348).

### DIFF
--- a/charts/crljanitor/templates/deployment.yaml
+++ b/charts/crljanitor/templates/deployment.yaml
@@ -53,7 +53,7 @@ spec:
         image: "{{- if .Values.image -}}
             {{ .Values.image }}
           {{- else -}}
-            {{ .Values.imageConfig.repository }}:{{ default .Values.appVersion .Values.imageConfig.tag }}
+            {{ .Values.imageConfig.repository }}:{{ default .Values.global.applicationVersion .Values.imageConfig.tag }}
           {{- end }}"
         imagePullPolicy: {{ .Values.imageConfig.imagePullPolicy }}
         {{- if not .Values.proxy.enabled }}
--- a/charts/crljanitor/templates/deployment.yaml
+++ b/charts/crljanitor/templates/deployment.yaml
@@ -17,11 +17,43 @@ spec:
     spec:
       serviceAccountName: terra-crl-janitor-service-sa
       containers:
+      - name: cloudsql-proxy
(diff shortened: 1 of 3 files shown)


charts/crljanitor/templates/deployment.yaml
```diff
@@ -17,11 +17,43 @@ spec:
...
+        volumeMounts:
+        - name: cloudsql-sa-creds
+          mountPath: /secrets/cloudsql
+          readOnly: true
       - name: terra-crl-janitor
         image: "{{- if .Values.image -}}
             {{ .Values.image }}
           {{- else -}}
-            {{ .Values.imageConfig.repository }}:{{ default .Values.global.applicationVersion .Values.imageConfig.tag }}
+            {{ .Values.imageConfig.repository }}:{{ default .Values.appVersion .Values.imageConfig.tag }}
```
choover-broad, line 56: this looks like an unintentional revert, it should use `.Values.global.applicationVersion`

charts/crljanitor/values.yaml
```diff
@@ -20,7 +20,13 @@ imageConfig:
   # @default -- The chart's appVersion value will be used
   tag:
   imagePullPolicy: Always
+
 replicas: 1
+
+
+# Allows CloudSQL proxy time to start up. See DDO-1352
```
choover-broad, line 27: nit: this description doesn't show up in the readme because it isn't prefixed with `# startupSleep --`

choover-broad (approved):
Should be good to merge once the image config thing is resolved. Thx!

charts/crljanitor/values.yaml
knapii-developments: Oh, that's why the syntax is like that? I had no idea! haha 6c8f739f

charts/crljanitor/templates/deployment.yaml
knapii-developments: 6c8f739f

knapii-developments:
> I'm not well versed in post start hooks and I think I haven't been following this all closely enough, how does making the proxy sleep after start prevent the app from starting up too early?

Placing `sleep` in the post start command prevents the next container in line, in this case, the application container, from starting until the command is complete.

PR #474, opened 2021-09-03
* DDO-1352 Add 30-second sleep for Thurloe couldSQL proxy

knapii-developments:


choover-broad (approved):
LGTM, thanks @knapii-developments!

mflinn-broad (approved):
Thanks Eno!

PR #478, opened 2021-09-04
[DDO-1444] Cluster release support

choover-broad:
Adapt our Argo App generator charts to support cluster releases.

### Changes
* Adapt the `terra-argocd-app` chart to support cluster releases
* Rename the `terra-argocd-env` chart to `terra-argocd-project`, because this chart will soon generate cluster release projects too
* Add a new label, `release`, to Terra ArgoCD apps. This is a more generically-named replacement of the existing `app` label, which doesn't make sense for some cluster charts like `terra-storage-classes`. (The `app` label will continue to exist for app releases, but not cluster releases)
* Add a new label, `jenkins-sync-group`, to Terra ArgoCD Apps. `jenkins-sync-group` will be used by Jenkins to determine which Argo apps to sync during the monolith release. It will be set to `prod` for Terra prod releases, `alpha` for alpha, etc. Currently, the `env` label is used for this purpose; I'm adding a dedicated separate label so that Jenkins can sync cluster releases during the monolith release as well
* Get rid of the shared templates chart, which added unnecessary complexity

### Related
* https://github.com/broadinstitute/terra-helmfile/pull/1638


charts/terra-argocd-app/templates/app.yaml
```diff
@@ -1,39 +1,45 @@
-{{- with .Values -}}
+{{- if .Values.releaseType | eq "app" }}
+# Generates an ArgoCD application for a Terra application release
+{{- $app := required "releaseName is required" .Values.releaseName }}
```
knapii-developments, line 3: Is there a reason we can't use `Release.Name` [link](https://helm.sh/docs/chart_template_guide/builtin_objects/)?

jack-r-warren (approved):
Awesome stuff, maybe down the road some of the bespoke dsp-ci stuff in the argocd chart can be made a little less bespoke

charts/terra-argocd-app/templates/app.yaml
```diff
@@ -1,39 +1,45 @@
...
+{{- $env := required "targetName is required" .Values.targetName }}
 apiVersion: argoproj.io/v1alpha1
 kind: Application
 metadata:
-  name: {{ required "A valid app name is required" .app }}-{{ required "A valid environment is required" .environment }}
+  name: "{{ required "A valid releaseName is required" .Values.releaseName }}-{{ required "A valid targetName is required" .Values.targetName }}"
```
knapii-developments, line 8: I'm a little confused here. On line 3, we're setting $app to .Values.releaseName. Can we not refer to $app here? I think it'll help with the readability and uniformity.

charts/terra-argocd-app/templates/app.yaml
```diff
@@ -1,39 +1,45 @@
...
   finalizers:
   - resources-finalizer.argocd.argoproj.io
   {{- end }}
 spec:
-  project: {{ include "terra-argocd.projectName" . }}
+  project: {{ required "A valid projectName is required" .Values.projectName | quote }}
   destination:
-    namespace: {{ include "terra-argocd.destinationNamespace" . }}
-    server: {{ quote .clusterAddress }}
+    namespace: {{ required "A valid namespace is required" .Values.namespace | quote }}
```
knapii-developments, line 27: Maybe `Release.Namespace` would be useful here?

charts/terra-argocd-app/templates/app.yaml
knapii-developments, line 3: Is there a reason we can't use `Release.Name` [link](https://helm.sh/docs/chart_template_guide/builtin_objects/)?

charts/terra-argocd-app/templates/legacy-configs-app.yaml
```diff
@@ -1,35 +1,40 @@
-{{- with .Values -}}
-{{- if .legacyConfigsEnabled -}}
+{{- if eq .Values.releaseType "app" | and .Values.legacyConfigsEnabled -}}
```
knapii-developments, line 1: I think you're trying to do a logical [and](https://helm.sh/docs/chart_template_guide/function_list/#and) here.

charts/terra-argocd-app/templates/cluster.yaml
```diff
@@ -0,0 +1,45 @@
+{{- if .Values.releaseType | eq "cluster" }}
+# Generates an ArgoCD application for a Terra cluster release
+{{- $release := required "releaseName is required" .Values.releaseName }}
```
knapii-developments, line 3: Same question here for Release.Name

charts/terra-argocd-app/templates/cluster.yaml
```diff
@@ -0,0 +1,45 @@
...
+{{- $cluster := required "targetName is required" .Values.targetName }}
+{{- if not ( .Values.targetName | eq .Values.clusterName ) }}
```
knapii-developments, line 5: `{{ if not (eq .Values.targetName .Values.clusterName) }}` I believe is what you're looking for here...? 
since $cluster is required, can we reference it here?

knapii-developments (changes requested):
I don't see any breaking changes. Just some minor helm questions/syntax suggestions. 👯

charts/terra-argocd-app/templates/app.yaml
choover-broad: The namespace that is used here (`terra-dev`, etc) is actually different from Release.Namespace (which is `ap-argocd`)

charts/terra-argocd-app/templates/app.yaml
choover-broad: `Release.Name` can't be used here. The `releaseName` value in this context refers to the name of the service that is being deployed by this ArgoCD application (eg. `cromwell`, `workspacemanager`, `diskmanager`). `Release.Name` would be the release name for this Helm chart ([eg. `terra-argocd-app-cromwell`](https://github.com/broadinstitute/terra-helmfile/blob/master/helmfile.yaml#L149))

charts/terra-argocd-app/templates/cluster.yaml
choover-broad: Can't be used unfortunately

charts/terra-argocd-app/templates/cluster.yaml
choover-broad: `eq .Values.targetName .Values.clusterName` is the same as `.Values.targetName | eq .Values.clusterName`, just a different style. I find the interleaved syntax more legible because of its similarity to other programming languages (`if a == b`) 🤷

charts/terra-argocd-app/templates/legacy-configs-app.yaml
choover-broad: I am, with pipeline syntax

charts/terra-argocd-app/templates/app.yaml
knapii-developments: Oh I see! That makes sense.

charts/terra-argocd-app/templates/app.yaml
knapii-developments: Thanks for explaining this!

charts/terra-argocd-app/templates/app.yaml
choover-broad: Adjusted!

charts/terra-argocd-app/templates/cluster.yaml
knapii-developments: I see. Here it was a little confusing because the output before the pipe is an input for after the pipe. I didn't know if you were trying to evaluate something prior to checking if the values are equal.

charts/terra-argocd-app/templates/legacy-configs-app.yaml
knapii-developments: OOOOO, I read this wrong.

knapii-developments (approved):

PR #485, opened 2021-09-15
* DDO-1428 Add PDB to Thurloe

knapii-developments:
Add [pod disruption budget ](https://v1-20.docs.kubernetes.io/docs/tasks/run-application/configure-pdb/)to reduce disruption during node pool upgrades This ensures minimum availability is maintained at all times

### DIFF
--- a/charts/thurloe/templates/podDisruptionBudget.yaml
+++ b/charts/thurloe/templates/podDisruptionBudget.yaml
@@ -3,7 +3,8 @@ kind: PodDisruptionBudget
 metadata:
   name: {{ .Values.name }}-pdb
 spec:
-  # don't reduce serving capacity by more than 30%
+  # 30% of pods must still be available after the eviction,
+  # even in the absence of the evicted pod.
   minAvailable: 30%
   selector:
     matchLabels:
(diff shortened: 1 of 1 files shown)


charts/thurloe/templates/podDisruptionBudget.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ .Values.name }}-pdb
+spec:
+  # don't reduce serving capacity by more than 30%
+  minAvailable: 30%
```
knapii-developments, line 7: I'm open to changing this number. Since the default replica is 3, I used 30% instead of 1. You can read about using percentages instead of numbers [here](https://v1-20.docs.kubernetes.io/docs/tasks/run-application/configure-pdb/#rounding-logic-when-specifying-percentages)

jack-r-warren: Maybe I'm just reading it wrong but the number seems in conflict with the comment--minimum available being 30% allows a reduction of up to 70%, but the comment says a reduction wouldn't exceed 30%

jack-r-warren (approved):

charts/thurloe/templates/podDisruptionBudget.yaml
knapii-developments: How do you feel about the new comment now?

jack-r-warren: looks great!

jack-r-warren (approved):

PR #498, opened 2021-09-17
Revert DDO-1530 / #489 This reverts commit cad91bff2ce272fcac4957a1d57f4a312f9ef2ad.

choover-broad:
Revert #489


em-may (approved):
If you have a minute to add some context and a plan for moving forward, that could be nice, especially if this gets shared around. Just an idea.

PR #502, opened 2021-09-21
Transfer Buffer service to use ingress library

knapii-developments:
Refactors Buffer's ingress, frontend config, backend config, and service charts to use the ingress library.


charts/buffer/templates/ingress/backendconfig.yaml
```diff
@@ -1,22 +1,3 @@
...
-    healthyThreshold: 2
-    unhealthyThreshold: 2
-    type: HTTPS
-    port: 443
-    requestPath: /status
-  {{- if .Values.ingress.securityPolicy }}
-  securityPolicy:
-    name: {{ .Values.ingress.securityPolicy }}
-  {{- end }}
+{{- include "libchart.backendconfig" (list . "buffer.backendconfig" )  -}}
```
jack-r-warren, line 1: what happens here if ingress is disabled and the library chart isn't included? What defines `libchart.backendconfig`?

knapii-developments: So the file now looks like this:
```
{{- define "buffer.ingress" -}}
{{- end }}
```
and in the values file `ingress.enabled: false`?

jack-r-warren: Sorry, let me rephrase:
What happens if ingress is disabled? The ingress chart wouldn't exist as a dependency, so line 1 of
```
{{- include "libchart.backendconfig" (list . "buffer.backendconfig" ) -}}
{{- define "buffer.backendconfig" -}}
{{- end }}
```
would be referencing a "libchart.backendconfig" that doesn't exist. What happens then?

jack-r-warren: Almost--I think we should put the `include` inside the `if`, not before it, because really what we want to say is "if ingress is disabled, don't even try to access the `libchart.backendconfig` it defines"

knapii-developments: The same issue applies [here](https://github.com/broadinstitute/terra-helm/pull/506#discussion_r716972208).

jack-r-warren: Ah, maybe include works differently. I'd always used `template` which does support being inside a conditional https://github.com/broadinstitute/terra-helm/blob/85d1cf7b62ef30423896759c87ec6f5f66113dce/charts/rawls/templates/migration.yaml

jack-r-warren (approved):

jyang-broad (approved):
looks sane! Just as a academic exercise, if we were to say, push a 0.7.0 version of the chart, would all the dependent charts still be able to run 0.6.0 until they're ready (or maybe a 1.0.0 for major breaking changes kind of thing)

charts/buffer/templates/ingress/backendconfig.yaml
knapii-developments: Yeah, they have different purposes. You can read more about it [here](https://helm.sh/docs/chart_template_guide/named_templates/#the-include-function). But in short: 

> Because template is an action, and not a function, there is no way to pass the output of a template call to other functions; the data is simply inserted inline.

PR #503, opened 2021-09-22
[DDO-1467] Liquibase *application* chart

jack-r-warren:
Make liquibase-migrations an application chart, with a rework of its values file and manifests.

- This PR contains the overall values file schema and the templates
	- Templates are made out of one file so I don't have to repeat looping/merging logic; my hope is that one file, readable top to bottom, is better than a bunch of indirection
- https://github.com/broadinstitute/terra-helm/pull/504 shows the impact on Leonardo and Rawls (legacy apps with simpler structure, this is what already worked) and Buffer (newer app, this has never worked before but works now)
- https://github.com/broadinstitute/terra-helmfile/pull/1708 shows the impact on env-specific values files

The best overview of this PR is probably to look at the [additions to Buffer's values file](https://github.com/broadinstitute/terra-helm/blob/fae44529ab86d0da82c09cdd0a26b05f76375fe3/charts/buffer/values.yaml#L187). There, we have no firecloud-develop like we do for Rawls/Leonardo--everything there is adapting either to choices made in Buffer's own chart design or in Buffer's codebase/compilation process.


charts/liquibase-migration/values.yaml
```diff
@@ -1,53 +1,106 @@
...
-#                                                                                     #
-#   DEMONSTRATION: This file is NOT used because this is a library chart.             #
-#   The source of truth is in the templates themselves, this file is documentation.   #
-#                                                                                     #
-#   All unset values are required.                                                    #
-#                                                                                     #
-# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+# -- Specific migrations to run; each config merged over the defaults to form the full configuration
+# @default -- None by default, one entry here required per migration
+enumerated:
```
choover-broad, line 3: nit: Call this `migrations`, `migrationJobs`, or `migrationDatabases` instead? `enumerated` feels a tad unclear 🤔

charts/liquibase-migration/values.yaml
```diff
@@ -1,53 +1,106 @@
...
+    secretName: null
+    # -- The full mount path of the file, ending with the file itself
+    mountFilePath: "/etc/liquibase.properties"
+  # -- Controls for setting up database authentication, one sub-object must be provided;
+  # existingKubernetesSecret takes precedence if provided
+  # @default -- None
+  migrationDatabaseCredentials:
+    # -- Use an existing Kubernetes secret for credentials directly
+    # @default -- With the secret name omitted, no existing Kubernetes secret will be accessed
+    existingKubernetesSecret:
```
choover-broad, line 84: nit: should this be `fromKubernetesSecret` for consistency with `fromVaultSecret`?

charts/liquibase-migration/values.yaml
```diff
@@ -1,53 +1,106 @@
...
-  # migration.labelRef -- (string) Template reference to `include` in resource labels
-  # @default -- Set to "${appName}.labels"
-  labelRef: null
-  # migration.syncWave -- (string) ArgoCD wave to apply these resources in
-  syncWave: "-1"
-  # migration.imageTag -- (string) Tag of the migration.imageName to use
-  # WARNING: App may behave unexpectedly if its database has been migrated on a different version than the app itself
+  k8sAppName: null
+  # -- (string) Template reference to use to obtain resource labels, set to empty to disable
+  # @default -- Set to `${k8sAppName}.labels`
```
choover-broad, line 19: Nice, love the use of the downstream chart template here!

charts/liquibase-migration/values.yaml
```diff
@@ -1,53 +1,106 @@
...
+    existingKubernetesSecret:
+      # -- (string) Name of the existing Kubernetes secret to use
+      name: null
+      # -- (string) Key of the username within the Kubernetes secret
+      usernameKey: null
+      # -- (string) Key of the password within the Kubernetes secret
+      passwordKey: null
+      # -- (string) Optional key of a database name in the Kubernetes secret, available as $DB_NAME in the migration container
+      # @default -- If null, $DB_NAME remains unset
+      variableDatabaseKey: null
```
choover-broad, line 93: nit: `databaseNameKey` for clarity?

charts/liquibase-migration/values.yaml
```diff
@@ -1,53 +1,106 @@
...
+  # @default -- Set to `${k8sAppName}.labels`
+  k8sLabelRef: null
+  # -- (string) App's Kubernetes service account (selective sync may be necessary if the KSA isn't applied before k8sSyncWave)
+  # @default -- Set to `${k8sAppName}-sa`
+  k8sServiceAccount: null
+  # -- (string) Argo CD wave to apply resources in
+  k8sSyncWave: "-1"
+
+  # -- Image to use for the Cloud SQL Proxy
+  proxyImage: "gcr.io/cloudsql-docker/gce-proxy"
```
choover-broad, line 28: nit: `sqlproxy` for clarity? We use `proxy` in many Helm charts to refer to the Apache proxy container, and `sqlproxy` for the CloudSQL proxy

charts/liquibase-migration/templates/manifests.yaml
```diff
@@ -0,0 +1,210 @@
...
+        image: {{ $config.migrationImage | required "migrationImage required" }}:{{ $config.migrationImageTag }}
+        {{- if (($config.migrationConfigFileMount).secretName) }}
+        volumeMounts:
+        - mountPath: {{ $config.migrationConfigFileMount.mountFilePath | required "migrationConfigFileMount.mountFilePath required if migrationConfigFileMount.secretName provided" }}
+          subPath: {{ $config.migrationConfigFileMount.mountFilePath | splitList "/" | last }}
+          name: migration-configs-volume
+          readOnly: true
+        {{- end }}
+        env:
+        - name: DB_USERNAME
```
choover-broad, line 144: nit: Any way to simplify some of this conditional logic, since it looks like it's repeated once for each environment variable?

Maybe populate a variable, $credentialsSecret, with the necessary attributes and use that for each variable?

choover-broad (approved):
This is great! 🎉 Left some nits in-line which you can feel free to ignore. :grin:

I love the use of an application chart with real default values. 💯

charts/liquibase-migration/values.yaml
jack-r-warren: it actually breaks in annoying ways, which is unfortunate. The root chart's template is re-rendered, which means any .Values reference in the template becomes that of liquibase-migrations. This means you have to do really unfortunate things like [this](https://github.com/broadinstitute/terra-helm/pull/504/files#diff-f0039169139fa8741996dc6770e9d2579a1574c3c0b6c2fc2a36db470af63af0R212) if the root chart uses .Values.name. A fix would be to stop using .Values.name and instead move things to .Values.global.name :/

charts/liquibase-migration/values.yaml
jack-r-warren: Good call, I'm going to rename everything in that series of variables, it helps wrangle helm-docs into producing useful output

charts/liquibase-migration/values.yaml
choover-broad: Ah, I guess that's the downside to the application chart approach. Can't pass in specific context like with a library chart.

I guess the library chart does still have some pros:
* can pass in templates by name, with any context at all
* can use merging to override rendered resources, as Eno demoed with the ingress lib chart

The goal with the application chart approach was to have a real `values.yaml` file with defaults right?
I'm wondering if we could simulate that in a library chart with some hacky trick like `$.Files.Get values.yaml | fromYaml`

(Just some rambling thoughts, not convinced either option is better)

PR #504, opened 2021-09-23
[DDO-1467] Liquibase chart downstream changes

jack-r-warren:
Downstream effects of #503. Check out Argo CD to see this working in rawls-dev, leonardo-dev, and buffer-dev courtesy of https://github.com/broadinstitute/terra-helmfile/pull/1708.

### DIFF
--- a/charts/buffer/README.md
+++ b/charts/buffer/README.md
@@ -13,7 +13,7 @@ Chart for Resource Buffering Service
 
 | Repository | Name | Version |
 |------------|------|---------|
-| https://broadinstitute.github.io/terra-helm/ | liquibase-migration | 1.0.0-503.1632773568.0fec2e |
+| https://terra-helm.storage.googleapis.com | liquibase-migration | 1.1.0 |
 
 ## Values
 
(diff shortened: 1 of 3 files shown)


charts/rawls/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: rawls
-version: 0.20.0
+version: 0.21.0
```
knapii-developments, line 3: Why did you bump this version? Doesn't chart releaser do this for us?

charts/rawls/values.yaml
```diff
@@ -101,24 +101,34 @@ deploymentDefaults:
...
+            name: "rawls-backend-sqlproxy-env"
+    migrationImage: "gcr.io/broad-dsp-gcr-public/rawls"
+    migrationArgsClasspath:
+      - "$(find /rawls -name 'rawls*.jar')"
+    migrationConfigFileMount:
+      secretName: "rawls-backend-app-ctmpls"
+    migrationDatabaseCredentials:
+      fromVaultSecret:
+        path: null
+        usernameKey: "slick_db_user"
```
knapii-developments, line 133: This makes me happy. haha

charts/rawls/Chart.yaml
jack-r-warren: I bumped it here to give me a fresh version number, I was having trouble getting Argo to respect the versions I was giving it through helmfile. I know that the chart releaser will bump it again

charts/rawls/values.yaml
jack-r-warren: yeah slick is the name of the database thing they use in the code, suppose it made its way to their vault config

PR #505, opened 2021-09-24
add ecm googleproject [POL-83]

Shakespeared:
Exposes a value that was already fed into the environment

### DIFF
--- a/charts/externalcreds/templates/deployment.yaml
+++ b/charts/externalcreds/templates/deployment.yaml
@@ -176,6 +176,8 @@ spec:
         {{- end }}
         - name: SAM_ADDRESS
           value: {{ $.Values.samAddress }}
+        - name: AUTHORIZATION_CHANGE_EVENTS_ENABLED
+          value: {{ $.Values.authorizationChangeEventsEnabled }}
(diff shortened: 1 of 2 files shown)


charts/externalcreds/templates/deployment.yaml
```diff
@@ -176,6 +176,8 @@ spec:
         {{- end }}
         - name: SAM_ADDRESS
           value: {{ $.Values.samAddress }}
+        - name: SERVICE_GOOGLE_PROJECT
+          value: { { $.Values.serviceGoogleProject } }
```
em-may, line 180:
```suggestion
          value: {{ $.Values.serviceGoogleProject }}
```

em-may (approved):
I'm curious to know how this value was already getting to the env.

charts/externalcreds/templates/deployment.yaml
Shakespeared: It wasn't actually getting through to the env, but was being defined in the helmfile - we just haven't needed it until now.

dvoet (approved):

PR #506, opened 2021-09-27
Add logic when ingress isn't enabled

knapii-developments:
This PR is to address this [comment](https://github.com/broadinstitute/terra-helm/pull/502/files#r714085722)

### DIFF
--- a/charts/ingresslib/templates/_backendconfig.yaml
+++ b/charts/ingresslib/templates/_backendconfig.yaml
@@ -1,4 +1,5 @@
 {{- define "libchart.backendconfig.tpl" -}}
+{{ if .Values.ingress.enabled -}}
 apiVersion: cloud.google.com/v1
 kind: BackendConfig
 metadata:
@@ -18,6 +19,7 @@ spec:
 {{- if .Values.ingress.securityPolicy }}
   securityPolicy:
     name: {{ .Values.ingress.securityPolicy }}
+{{- end }}
 {{- end }}
 {{- end }}
 
--- a/charts/ingresslib/templates/_frontendconfig.yaml
+++ b/charts/ingresslib/templates/_frontendconfig.yaml
@@ -1,4 +1,5 @@
 {{- define "libchart.frontendconfig.tpl" -}}
+{{ if .Values.ingress.enabled -}}
 apiVersion: networking.gke.io/v1beta1
 kind: FrontendConfig
 metadata:
@@ -9,6 +10,7 @@ metadata:
 spec:
   sslPolicy: {{ .Values.ingress.sslPolicy }}
 {{- end }}
+{{- end -}}
 {{- end -}}
 
 {{- define "libchart.frontendconfig" -}}
--- a/charts/ingresslib/templates/_ingress.yaml
+++ b/charts/ingresslib/templates/_ingress.yaml
@@ -1,4 +1,5 @@
 {{- define "libchart.ingress.tpl" -}}
+{{ if .Values.ingress.enabled -}}
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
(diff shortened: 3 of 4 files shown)


jack-r-warren:
This is good and needed, I'm sorry for not giving more of an explanation back at the [start of this discussion](https://github.com/broadinstitute/terra-helm/pull/422#discussion_r675715879). I think my point from that original comment and the one you reference still stands: right now, the "call site" looks like this--
```
{{- include "libchart.backendconfig" (list . "buffer.backendconfig" )  -}}
{{- define "buffer.backendconfig" -}}
{{- end }}
```
The fact that there's not an if statement around the call to the `libchart.backendconfig` that may-or-may-not exist is part of what caused this issue.

I think we should have if statements in both locations: one around the outside for the "semantically correct, if-we-don't-import-it-we-can't-reference-it" reason, and one on the inside in the chart like you have here for the "helm actually has weird behavior and it'll reference unimported charts if asked to" backup reason.

knapii-developments:
> This is good and needed, I'm sorry for not giving more of an explanation back at the [start of this discussion](https://github.com/broadinstitute/terra-helm/pull/422#discussion_r675715879). I think my point from that original comment and the one you reference still stands: right now, the "call site" looks like this--
> 
> ```
> {{- include "libchart.backendconfig" (list . "buffer.backendconfig" )  -}}
> {{- define "buffer.backendconfig" -}}
> {{- end }}
> ```
> 
> The fact that there's not an if statement around the call to the `libchart.backendconfig` that may-or-may-not exist is part of what caused this issue.
> 
> I think we should have if statements in both locations: one around the outside for the "semantically correct, if-we-don't-import-it-we-can't-reference-it" reason, and one on the inside in the chart like you have here for the "helm actually has weird behavior and it'll reference unimported charts if asked to" backup reason.

Thanks for explaining this further. I updated the [buffer ingress chart PR](https://github.com/broadinstitute/terra-helm/pull/502/files) to ensure we're on the same page

charts/ingresslib/templates/_backendconfig.yaml
```diff
@@ -1,4 +1,5 @@
 {{- define "libchart.backendconfig.tpl" -}}
+{{ if .Values.ingress.enabled -}}
```
jack-r-warren, line 2: I left the same comment on the other PR-if ingress is disabled, I don't think we want to even `define` anything here, so I don't think the define should come before the `if` (we don't want an unenabled chart to still have side effects, like adding a bunch of empty variables)

knapii-developments: I don't think so. Because [this line](https://github.com/broadinstitute/terra-helm/blob/48f9ad51b99931f0875189d9561a2f4330390b1c/charts/ingresslib/templates/_backendconfig.yaml#L27) expects ```libchart.backendconfig.tpl``` template to be defined. If I place the if statement before the define, the function may not exist which would cause a template not defined error. An empty template is okay here. Because ```libchart.util.merge```  takes common values between the templates and overrides the values seen here. Also, it adds any additional information.

jack-r-warren: I guess what I'm saying is that literally the entire file should be wrapped in these "if ingress enabled" statements. If the chart isn't enabled, everything here should do absolutely nothing. No references, nothing gets defined.

I agree that an empty template technically works here, but right now "disabled" means "well it still does things but not anything useful" instead of "it does nothing"

knapii-developments: I don't think you can place if statements outside of a template definition. I tried rendering the chart when the template definition is  wrapped in the if statement, as well as putting the entire file in an if statement and I kept getting the error ```unexpected <define> in command```. I guess this kinda makes sense. In general, I would try not to wrap a function/method/class in an if statement. I wouldn't do something like:

```
if blah:
   def my_function():
      ...
```
or 
```
if blah:
   class ClassDef
```

jack-r-warren: Ah I gotcha, I just looked at what I'd done in the part and you're totally correct, sorry for sending you on a wild goose chase. This looks great, and thank you for the explanation!

jack-r-warren (approved):

PR #507, opened 2021-09-27
* DDO-1567 Add PDB library

knapii-developments:
Adds pod disruption library for services to utilize

### DIFF
--- a/charts/pdblib/Chart.yaml
+++ b/charts/pdblib/Chart.yaml
@@ -1,0 +1,11 @@
+apiVersion: v2
+name: pdb-lib
+description: A libary Helm chart for Pod Disruption Budgets
+type: library
+version: 0.0.0
+keywords:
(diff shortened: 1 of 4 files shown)


charts/pdblib/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+name: pdb-lib
+description: A libary Helm chart for Pod Disruption Budgets
+type: library
+version: 0.0.0
```
em-may, line 5: Do we usually start at `0.0.0` or `0.0.1`?

em-may (approved):

charts/pdblib/Chart.yaml
knapii-developments: I started with 0.0.0 because the chart releaser will bump it to 0.1.0

PR #511, opened 2021-09-28
[DDO-1467] Leonardo Chart: use newer liquibase-migration subchart

jack-r-warren:
The liquibase-migration chart got a big update in #503 to support apps following Buffer's config/deployment model. 

The end effect is the same, but the configuration at the "call-site" looks very different. The liquibase-migration chart is now an application chart instead of a library chart. This changes how charts like Leonardo's import and use it, but means that the liquibase-migration chart can now properly have its own [documentation and default values](https://github.com/broadinstitute/terra-helm/blob/master/charts/liquibase-migration/values.yaml).

See https://github.com/broadinstitute/terra-helmfile/pull/1728 for the environment-specific values file changes.


rtitle (approved):
LGTM

Qi77Qi (approved):

choover-broad (approved):

PR #513, opened 2021-09-28
[DDO-1467] Buffer Chart: Add liquibase-migration subchart

jack-r-warren:
The liquibase-migration helper chart has been reworked and has been tested in dev. This PR adds it to Buffer's chart and the accompanying terra-helmfile PR (https://github.com/broadinstitute/terra-helmfile/pull/1730) will enable the migrations as the updated chart propagates through the release cadence.

For background:
When apps start and block their startup on running possibly-lengthy Liquibase migrations, it is harder to detect and fix actual issues with app startup (in other words, hour-long startups are within the realm of expected behavior, so automated checks need to not fire). 

Our solution is to _also_ run an app's migrations before the app ever starts, using an Argo CD sync hook. We use the app image's own changelogs, liquibase version, JDBC driver, etc. When the app starts, Liquibase simply exits quickly because no migration is necessary. No changes are needed to the app's codebase, it is a free speed boost to startup where enabled. We've done this for Rawls and Leonardo and we've now adapted our helper chart to work for apps like Buffer.

Here, this PR adds this new pre-migration feature for both Buffer's own database and Buffer's stairway database. Both migrations run simultaneously on separate pods with their own Cloud SQL proxies. Any issues would cause errors during Argo CD sync during "sync wave -1", meaning before literally anything else has been touched.

### DIFF
--- a/charts/buffer/values.yaml
+++ b/charts/buffer/values.yaml
@@ -183,3 +183,67 @@ buffer:
 
 # cleanupAfterHandout -- Whether to publish message to Janitor after resource is handed out.
 cleanupAfterHandout: false
+
+liquibase-migration:
+  # In an env-specific file use something like the following to fully enable:
+  # ```
+  # liquibase-migration:   
+  #   defaults:
+  #     enabled: true
+  # ```
+  # See https://github.com/broadinstitute/terra-helm/blob/master/charts/liquibase-migration for more information
+
+  migrationJobs:
+  - name: buffer
+    # on filesystem, on classpath via app/resources/
+    migrationArgsConfigChangelog: "db/changelog.xml"
+    migrationDatabaseCredentials:
+      fromKubernetesSecret:
+        name: "buffer-postgres-db-creds"
+  - name: buffer-stairway
+    # inside stairway jar, on classpath via app/libs/*
+    migrationArgsConfigChangelog: "stairway/db/changelog.xml"
+    migrationDatabaseCredentials:
+      fromKubernetesSecret:
+        name: "buffer-stairway-db-creds"
+
+  # Template definition buffer.labels attempts to read .Values.name from within liquibase-migration.
+  # Either we must pass the name again, move it to globals, or disable liquibase-migration labels.
+  name: "buffer"
+
+  defaults:
+    enabled: false
+    k8sServiceAccount: "buffer-service-sa"
+    sqlproxyCredentialFileMount:
+      secretName: "buffer-cloudsql-sa"
+      mountFilePath: "/secrets/cloudsql/service-account.json"
+    sqlproxyArgsInstances: "$(SQL_INSTANCE_PROJECT):$(SQL_INSTANCE_REGION):$(SQL_INSTANCE_NAME)=tcp:5432"
+    sqlproxyContainerConfig:
+      env:
+        - name: "SQL_INSTANCE_PROJECT"
+          valueFrom:
+            secretKeyRef:
+              name: "buffer-cloudsql-postgres-instance"
+              key: "project"
+        - name: "SQL_INSTANCE_REGION"
+          valueFrom:
+            secretKeyRef:
+              name: "buffer-cloudsql-postgres-instance"
+              key: "region"
+        - name: "SQL_INSTANCE_NAME"
+          valueFrom:
+            secretKeyRef:
+              name: "buffer-cloudsql-postgres-instance"
+              key: "name"
+    migrationImage: "gcr.io/terra-kernel-k8s/terra-resource-buffer"
+    migrationArgsClasspath:
+      - "app/libs/*"
+      - "app/resources/"
+    migrationArgsConfigDriver: "org.postgresql.Driver"
+    migrationArgsConfigUrl: "jdbc:postgresql://127.0.0.1:5432/${DB_NAME}"
+    migrationDatabaseCredentials:
+      fromKubernetesSecret:
+        usernameKey: "username"
+        passwordKey: "password"
+        databaseNameKey: "db"
+
--- a/charts/buffer/Chart.yaml
+++ b/charts/buffer/Chart.yaml
@@ -11,3 +11,7 @@ keywords:
 sources:
   - https://github.com/broadinstitute/terra-helm/tree/master/charts
   - https://github.com/DataBiosphere/terra-resource-buffer
+dependencies:
+  - name: liquibase-migration
+    version: 1.1.0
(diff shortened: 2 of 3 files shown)


charts/buffer/values.yaml
```diff
@@ -183,3 +183,67 @@ buffer:
...
+    migrationArgsConfigChangelog: "db/changelog.xml"
+    migrationDatabaseCredentials:
+      fromKubernetesSecret:
+        name: "buffer-postgres-db-creds"
+  - name: buffer-stairway
+    # inside stairway jar, on classpath via app/libs/*
+    migrationArgsConfigChangelog: "stairway/db/changelog.xml"
+    migrationDatabaseCredentials:
+      fromKubernetesSecret:
+        name: "buffer-stairway-db-creds"
```
yonghaoy, line 208: nit, Buffer and other apps are using` {{ .Values.name }} `not `buffer`
For example: `{{ .Values.name }}-postgres-db-creds`
https://github.com/broadinstitute/terra-helm/blob/master/charts/buffer/templates/secrets.yaml#L5

charts/buffer/values.yaml
```diff
@@ -183,3 +183,67 @@ buffer:
...
+          valueFrom:
+            secretKeyRef:
+              name: "buffer-cloudsql-postgres-instance"
+              key: "region"
+        - name: "SQL_INSTANCE_NAME"
+          valueFrom:
+            secretKeyRef:
+              name: "buffer-cloudsql-postgres-instance"
+              key: "name"
+    migrationImage: "gcr.io/terra-kernel-k8s/terra-resource-buffer"
```
yonghaoy, line 238: Can this be the same as: 
https://github.com/broadinstitute/terra-helm/blob/master/charts/buffer/templates/deployment.yaml#L93L98?

charts/buffer/values.yaml
```diff
@@ -183,3 +183,67 @@ buffer:
...
+    migrationArgsClasspath:
+      - "app/libs/*"
+      - "app/resources/"
+    migrationArgsConfigDriver: "org.postgresql.Driver"
+    migrationArgsConfigUrl: "jdbc:postgresql://127.0.0.1:5432/${DB_NAME}"
```
yonghaoy, line 243: what is the DB_NAME?  Currently we have two: https://github.com/broadinstitute/terra-helm/blob/master/charts/buffer/templates/deployment.yaml#L117
And https://github.com/broadinstitute/terra-helm/blob/master/charts/buffer/templates/deployment.yaml#L134

charts/buffer/values.yaml
```diff
@@ -183,3 +183,67 @@ buffer:
...
+  migrationJobs:
+  - name: buffer
+    # on filesystem, on classpath via app/resources/
+    migrationArgsConfigChangelog: "db/changelog.xml"
+    migrationDatabaseCredentials:
+      fromKubernetesSecret:
+        name: "buffer-postgres-db-creds"
+  - name: buffer-stairway
+    # inside stairway jar, on classpath via app/libs/*
+    migrationArgsConfigChangelog: "stairway/db/changelog.xml"
```
yonghaoy, line 205: @ddietterich checking Stairway DB migration path

charts/buffer/values.yaml
jack-r-warren: This sort of extreme parameterization isn't something we can support well. The context in an application subchart doesn't point to the same `.Values` that we have access to at the root: the other `.Values.liquibase-migration` is visible to the sub-chart just as `.Values`, so things like `.Values.name` or `.Values.image` at the root aren't available.

We do have a fix: `.Values.global.*` is special, and _is_ available. We could move common "global for a chart" values there to give us more options. However, that's standardization we'd want to do across all charts, which is why we're effectively hardcoding the name for now. We are only enabling this for the live environments, and there the `.Values.name` always equals "buffer".

As a note, we are actually already having to account for this issue with this [very bad hack](https://github.com/broadinstitute/terra-helm/pull/513/files/16e833dcbd54eebf1c4b90577b1a9401d0d1feab#diff-f0039169139fa8741996dc6770e9d2579a1574c3c0b6c2fc2a36db470af63af0R212), and before we add `tpl` calls to render random strings (and encourage this pattern) we'd like to have the global thing sorted out

charts/buffer/values.yaml
jack-r-warren: My above comment mostly still applies here I think, but this is actually an even more complex case because of the conditional logic. The underlying chart standardizes to Argo CD--Argo CD is what knows to actually run the migration based on annotations, and the migration expects application versions the standard way that Argo CD passes it.

I totally acknowledge that this won't work for Skaffold or other non-terra-helmfile deployments--we have no intention of enabling this there. We're trying to target the union of "all Terra apps deployed via terra-helmfile", so we need to be extremely careful about what per-app logic we put in the chart. Many apps have quirks about overriding image tags etc, but we've tried to standardize to how Argo CD and our rollback/release processes do it to find the common denominator.

charts/buffer/values.yaml
jack-r-warren: It is read from the same k8s secret that stores the database username and password. The option is documented [here](https://github.com/broadinstitute/terra-helm/blob/master/charts/liquibase-migration/values.yaml#L91), we set it [here](https://github.com/broadinstitute/terra-helm/pull/513/files/16e833dcbd54eebf1c4b90577b1a9401d0d1feab#diff-f0039169139fa8741996dc6770e9d2579a1574c3c0b6c2fc2a36db470af63af0R248).

charts/buffer/values.yaml
jack-r-warren: The path is hardcoded [here](https://github.com/DataBiosphere/stairway/blob/90755b0c79104a2d193487a89687ac8255487bf7/stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java#L152), from what I could find. When I ran this in dev it was able to find the changelogs and acknowledged that no changes were required to the stairway db it was connected to.

yonghaoy: Ah, I see. I tagged Dan because I know he is making some changes on Stairway, and make it imported from TCL(terra-common-lib): https://github.com/DataBiosphere/terra-resource-buffer/pull/189
But I guess that part remains unchanged.

yonghaoy (approved):

choover-broad (approved):
LGTM -- looking forward to seeing this in action on a service that uses stairway!

PR #519, opened 2021-09-29
* DDO-1578 Add pdb to consent

knapii-developments:


charts/consent/values.yaml
```diff
@@ -100,7 +100,7 @@ proxyLogLevel:
 proxyImageRepository:
 proxyImageVersion:
 
-replicas: 1
+replicas: 3
```
em-may, line 103: Does this cause any downtime? Mostly curious as I don't know much about how replicas are handled.

em-may (approved):

PR #520, opened 2021-09-29
[DDO-1546] Remove network policy config from Revere's chart

jack-r-warren:
An outcome of the meeting with Denis (DDO-1516) is that we don't need to worry about GKE sandbox (so no need to customize runtime class, but that's minor) or network policies (the big one). I close out my stale terra-helm-definitions draft PR that had all the network policy IPs in it, and this PR removes the templates and values defaults for those unneeded features.

This isn't deployed anywhere still, not really anything to test.


choover-broad (approved):
Really appreciate all the attention to maintainability here, @jack-r-warren. Thanks!

mflinn-broad (approved):

jyang-broad (approved):
looks sane!

PR #525, opened 2021-10-01
[No ticket]Refactor PDB to accept match labels from inheriting charts.

knapii-developments:
PR allows match labels defined in inheriting charts and checks to make sure they're defined.

### DIFF
--- a/charts/pdblib/templates/_util.yaml
+++ b/charts/pdblib/templates/_util.yaml
@@ -1,0 +1,14 @@
+{{- /*
+libchart.util.merge will merge two YAML templates and output the result.
+This takes an array of three values:
+- the top context
+- the template name of the overrides (destination)
+- the template name of the base (source)
+*/}}
+
+{{- define "pdblib.util.merge" -}}
+{{- $top := first . -}}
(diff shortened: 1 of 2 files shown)


charts/pdblib/templates/_util.yaml
```diff
@@ -0,0 +1,18 @@
...
+An error will be thrown if the top context does not provide values for
+.spec.selector.matchLabels
+*/}}
+
+{{- define "pdblib.util.merge" -}}
+{{- $top := first . -}}
+{{- $overrides := fromYaml (include (index . 1) $top) | default (dict ) -}}
+{{- $tpl := fromYaml (include (index . 2) $top) | default (dict ) -}}
+{{- $mergedmanifest := merge $overrides $tpl -}}
+{{ required "spec.selector.matchLabels are required." $mergedmanifest.spec.selector.matchLabels }}
```
jack-r-warren, line 16: I think this will be printed to console upon execution, but I'm not positive for definitions. Since you're executing this just for effect you could trap any output:
```suggestion
{{ $_ := required "spec.selector.matchLabels are required." $mergedmanifest.spec.selector.matchLabels }}
```

jack-r-warren (approved):

PR #529, opened 2021-10-05
QA-1595: Helm Charts for TestRunner Dashboard service.

ichengchang:
This PR is intended to set up the Kubernetes resources for running the `TestRunner Dashboard` in a cluster.

**Details of the architecture and minimum requirements**

- HTTPS traffic is enforced between the outside world and the native HTTP(S) Google Cloud Load Balancer (configured by ingress annotation `networking.gke.io/v1beta1.FrontendConfig`)
- Backend is a python-based http server (`Flask`). The traffic from Load Balancer to backend is currently unencrypted.
- The dashboard needs to be accessible through a DNS name.
- The dashboard use workload identity to enable KSA to act as certain GSA within same project. The GSA can be tailored to give access to GCP resources in multiple projects.


charts/testrunnerdashboard/templates/workloadidentity.yaml
```diff
@@ -0,0 +1,28 @@
+# This step requires Config Connector to be installed in the cluster.
+# The equivalent terraform-ap-modules/testrunner/dashboard.tf can be used
+# in place of this manifest if Config Connector is not enabled.
+{{- if .Values.secrets.gcpServiceAccount.workloadIdentity.enabled }}
```
jack-r-warren, line 4: I am working on config connector for the new cluster, but not in a way that will support project-wide IAM just yet (that requires the most dangerous permissions on config connector itself, so we're not there yet). Let's go with the Terraform option for using workload identity (which I have specially enabled there), we do that a number of other places in our devops infra

charts/testrunnerdashboard/templates/ingress/service.yaml
```diff
@@ -0,0 +1,23 @@
+{{ if .Values.ingress.enabled -}}
```
jack-r-warren, line 1: @knapii-developments what are your thoughts here? Is this something where we should start off using the ingress library you've been working on from the start (there's a few other ingress files here, too)?

charts/testrunnerdashboard/templates/_labels.tpl
```diff
@@ -0,0 +1,14 @@
...
+See
+https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
+*/ -}}
+{{- define "testrunnerdashboard.labels" -}}
+helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+app.kubernetes.io/name: {{ .Chart.Name }}
+app.kubernetes.io/instance: {{ .Values.name | quote }}
+app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+app.kubernetes.io/component: testrunnerdashboard
+app.kubernetes.io/part-of: terra
```
jack-r-warren, line 13: there's a chance that this specifically might change. We're not planning on deploying this to a Terra environment--we'll be deploying it to a new cluster, dsp-eng-tools. `part-of: tools` might make more sense here

knapii-developments (changes requested):
This is great helm work! 

We want app teams to use library charts from ingress, frontend configurations, backend configurations, and service charts. The library provides consistency and enforces some standards we have. 

Here is an example of [buffer using the ingress library ](https://github.com/broadinstitute/terra-helm/tree/master/charts/buffer/templates/ingress). It imports the dependency in [Chart.yaml](https://github.com/broadinstitute/terra-helm/blob/ddfb806a6a38568fb38833289772e19be90e3489/charts/buffer/Chart.yaml#L15). Notice in [backendconfig.yaml](https://github.com/broadinstitute/terra-helm/blob/master/charts/buffer/templates/ingress/backendconfig.yaml), it shows that it's possible to overwrite the values in the library [backendconfig.yaml ](https://github.com/broadinstitute/terra-helm/blob/master/charts/ingresslib/templates/_backendconfig.yaml). If you have any question, please let me know!

charts/testrunnerdashboard/values.yaml
```diff
@@ -0,0 +1,141 @@
...
+  # -- (string) Image repository
+  repository: us-central1-docker.pkg.dev/dsp-artifact-registry/terra-test-runner-dashboard/terra-test-runner-dashboard
+  # -- (string) Image tag
+  # @default -- global.applicationVersion
+  tag: null
+  # -- (string) When to pull images
+  imagePullPolicy: Always
+
+# -- (number) Number of replicas for the deployment
+replicas: 0
```
knapii-developments, line 87: Our standard is 3. Can you help me understand why you chose 0?

charts/testrunnerdashboard/templates/ingress/service.yaml
ichengchang: @jack-r-warren @knapii-developments The ingress has been refactored to use `ingresslib`. Please have a look and let me know how it goes. Thanks!

charts/testrunnerdashboard/templates/deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+        {{- if .Values.probes.liveness.enabled }}
+        livenessProbe:
+          {{- toYaml .Values.probes.liveness.spec | nindent 10 }}
+        {{- end }}
+        {{- if $.Values.probes.startup.enabled }}
+        startupProbe:
+          {{- toYaml $.Values.probes.startup.spec | nindent 10 }}
+        {{- end }}
+        command: ["python",
+                  "dashboard.py"]
```
jack-r-warren, line 50: Nit, let's avoid multiline json lists in yaml. Either use this:

```yaml
	command: ["python", "dashboard.py"]
```

or this:

```yaml
        command:
	  - python
	  - dashboard.py
```

charts/testrunnerdashboard/templates/service-account.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Values.name }}-service-ksa
```
jack-r-warren, line 4: We typically standardize to `{{ .Values.name }}-sa` for k8s SA names

jack-r-warren (changes requested):

charts/testrunnerdashboard/templates/deployment.yaml
ichengchang: Updated to yaml list syntax.

charts/testrunnerdashboard/templates/service-account.yaml
ichengchang: Updated k8s service account name refs in all related manifests.

jack-r-warren (approved):
Approving to unblock, I haven't taken a close look

choover-broad (approved):
Approving to unblock without actually reviewing, since this service won't be deployed to prod.

knapii-developments:
I made one comment about the recent update to the repository the Ingress Library lives in. Otherwise, the Backend,Frontend configs, service, and ingress charts look great!

ichengchang:
Replaced some of the `{{ .Values.name }}` by `{{ .Chart.Name }}` to match the `ingresslib` usage.

PR #531, opened 2021-10-06
Update name in ingress and pdb library

knapii-developments:
Name should be .Chart.name.


mflinn-broad (commented):
Does this end up using the name of the ingress library chart or the name of the parent chart using the library?

PR #539, opened 2021-10-12
Add PDB to externalCreds

knapii-developments:

### DIFF
--- a/charts/externalcreds/templates/podDisruptionBudget.yaml
+++ b/charts/externalcreds/templates/podDisruptionBudget.yaml
@@ -1,0 +1,8 @@
+{{- include "libchart.pdb" (list . "externalcreds.pdb" ) }}
+
+{{- define "externalcreds.pdb" -}}
+spec:
+  selector:
+    matchLabels:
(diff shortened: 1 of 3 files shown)


charts/externalcreds/templates/podDisruptionBudget.yaml
```diff
@@ -0,0 +1,8 @@
+{{- include "libchart.pdb" (list . "externalcreds.pdb" ) }}
+
+{{- define "externalcreds.pdb" -}}
+spec:
+  selector:
+    matchLabels:
+      deployment: {{ $.Values.name }}-{{ .name }}-deployment
```
choover-broad, line 7: I'm not sure this line will render correctly with the way the chart is currently set up (two deployments)

choover-broad (approved):
I think we need two PDBs, one for each deployment? Not sure this will work as expected as-is

charts/externalcreds/templates/podDisruptionBudget.yaml
knapii-developments: a747ad95

PR #548, opened 2021-10-12
*DDO-1468 Refactor Cromiam to use Ingress Library

knapii-developments:

### DIFF
--- a/charts/cromiam/templates/ingress/service.yaml
+++ b/charts/cromiam/templates/ingress/service.yaml
@@ -1,4 +1,7 @@
 {{- include "libchart.service" (list . "cromiam.service" )}}
 
 {{- define "cromiam.service" -}}
+spec:
+  selector:
+    deployment: {{ .Chart.Name }}-deployment
 {{- end }}


charts/cromiam/templates/ingress/backendconfig.yaml
```diff
@@ -1,22 +1,4 @@
...
-    checkIntervalSec: 5
-    timeoutSec: 5
-    healthyThreshold: 2
-    unhealthyThreshold: 2
-    type: HTTPS
-    port: 443
-    requestPath: /engine/latest/version
-{{- if .Values.ingress.securityPolicy }}
-  securityPolicy:
-    name: {{ .Values.ingress.securityPolicy }}
```
mflinn-broad, line 20: Cromiam is a private and service and is behind a security policy firewall. I just want to make sure you've accounted for that case in the ingress library?

knapii-developments: Oooh! Good question and I have [here](https://github.com/broadinstitute/terra-helm/blob/d667b303a4589fffaa2b628bdcffb350178ada45/charts/ingresslib/templates/_backendconfig.yaml#L19)

mflinn-broad (approved):

PR #561, opened 2021-10-15
DDO-1601 Sherlock Initial Helm Chart

mflinn-broad:

### DIFF
--- a/charts/sherlock/Chart.yaml
+++ b/charts/sherlock/Chart.yaml
@@ -1,0 +1,13 @@
+apiVersion: v2
+name: sherlock
+version: 0.0.0
+description: Chart for Sherlock - DSP's environment tracking and management service
+type: application
+keywords:
+  - dsp
+  - broadinstitute
+  - terra
+  - sherlock
+sources:
+  - https://github.com/broadinstitute/terra-helm/tree/master/charts
+  - https://github.com/broadinstitute/sherlock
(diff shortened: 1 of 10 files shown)


charts/sherlock/templates/ingress/cert.yaml
```diff
@@ -0,0 +1,21 @@
...
+    {{- include "sherlock.labels" . | nindent 4 }}
+spec:
+  dnsNames: 
+  - "{{ .Values.ingress.domain.hostname }}.{{ .Values.ingress.domain.suffix }}"
+  duration: 2160h0m0s
+  issuerRef:
+    group: cert-manager.io
+    kind: ClusterIssuer
+    name: letsencrypt-prod
+  renewBefore: 720h0m0s
```
em-may, line 16: Any way to specify days instead?

charts/sherlock/templates/ingress/backendConfig.yaml
```diff
@@ -0,0 +1,22 @@
+{{ if .Values.ingress.enabled -}}
```
jyang-broad, line 1: Can we use Eno's ingress libchart here?

charts/sherlock/Chart.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v2
+name: sherlock
+version: 0.0.0
```
jyang-broad, line 3: er does version 0.0.0 work or does it need to be 0.1.0 or something non zero

mflinn-broad: It'll get auto-bumped bumped to 0.1.0 once I merge this :)

PR #612, opened 2021-10-29
Actually remove postgres refs

em-may:
This PR correctly removes references to the postgres chart elsewhere so that it can be revamped for preview envs. The externalcreds and workspacemanager charts render correctly, which wasn't checked previously.


mflinn-broad (approved):
lgtm

jyang-broad (approved):
Looks Sane! 
No-op Comments: 
Looks like there's a change to the repos for pdb-lib and ingresslib here, confirmed they're intentional.
Also looks like theres updates to the autodocs prob has some old changes bundled in.
Did a grep to check for postgres.enabled and found nothing so looks good!

PR #619, opened 2021-11-02
[No-ticket4] Change opendj restart cronjob start stime

knapii-developments:
Changes cronjob time to 6 AM EDT and adds enabled value to values.yaml


choover-broad (approved):
Looks great Eno, you beat me haha!

thank you!

let's leave this disabled for now, we can always re-enable if the issue recurs next month.

PR #620, opened 2021-11-03
[DDO-1517] Revere chart

jack-r-warren:
Grab bag of changes to get this from what I wrote months ago to working.
- Add a job to run before the deployment executes
- Delay deployment execution by putting it in a later sync wave
- Add node tolerations and stuff so that we can point this at workload identity nodes that Mike set up in dsp-tools
- Don't use config connector at all because we're doing workload identity via Terraform
- Add a restart policy to the job


johnmarc (approved):

charts/revere/templates/job.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Values.name }}-job
```
mflinn-broad, line 4: I think this could also be done via an `initContainer` rather than an entire separate job that spins up new pod. No big deal though, just another option.

mflinn-broad (approved):
Good stuff, lgtm

charts/revere/templates/job.yaml
jack-r-warren: actually very true, ive just written more job yamls than initContainer yamls 🤷  this is an artifact from when we were thinking we were going to have multiple revere replicas--then a job to only run `prepare` once makes more sense. I'm going to leave it as-is for now as one less thing to change if we move to multiple replicas in the future

PR #623, opened 2021-11-08
[DDO-1653] Generate docs everywhere

jack-r-warren:
Recursive helm-docs + add a goptmpl for our special argocd docs so we don't lose them + ignore the prometheus operator chart because it has special docs, is deprecated, and no longer builds if we were to update it


choover-broad (approved):
thanks for this

PR #626, opened 2021-11-10
[DDO-1726] Add dsp-grafana

jack-r-warren:
`grafana/grafana` _can_ make an Ingress resource, but I couldn't get it working at all and it always had a different pattern (both in-yaml and in GCP UI) than all our other ingresses. And I need to do the frontend and backend configs anyway, so this just does manual ingress referencing the automatically-made service resource.

I deployed this Wednesday as part of https://github.com/broadinstitute/terra-helmfile/pull/1963 via https://github.com/broadinstitute/terraform-ap-deployments/pull/466


choover-broad (approved):
Rubber stamp

knapii-developments:
Is there a reason we're not using the ingress library for this?

jack-r-warren:
> Is there a reason we're not using the ingress library for this?

@knapii-developments the ingress library very much expects the connection between the app and load balancer to be HTTPS, but Grafana doesn't terminate TLS itself so that doesn't work here. My customization of the ingress library would be so sweeping/opaque as to be more brittle than just declaring directly what I want. I'm following the pattern of Sherlock and ArgoCD here more than I am a standard app.

PR #638, opened 2021-11-17
DDO-1754 Helm chart for deploying primary thanos components

mflinn-broad:
This is a first pass at a chart for deploying the primary thanos components to dsp-tools. It is a thin wrapper around the `bitnami/thanos` chart which adds custom ingress configuration. In this iteration only the `thanos query` component is deployed which is what performs aggregation and deduplication across multiple prometheuses. The other thanos components that handle compaction and long term storage will be added in a future iteration. 

Associated helmfile pr: https://github.com/broadinstitute/terra-helmfile/pull/1990


choover-broad (approved):
lgtm

PR #639, opened 2021-11-18
[DDO-1726] Fix a runner typo

jack-r-warren:
Would help if it knew that weekends were recurring. This scales down runners on the weekends to save money but it only did it once.

### DIFF
--- a/charts/dsp-argocd/values.yaml
+++ b/charts/dsp-argocd/values.yaml
@@ -131,4 +131,4 @@ apps:
       version: 0.2.0
     instances:
       chart: dsp-gha-runner-instances
-      version: 0.5.0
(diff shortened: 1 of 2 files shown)


choover-broad (approved):

jyang-broad (approved):
looks sane, just a version bump and a weekly recurrance for some job.

jack-r-warren:
> looks sane, just a version bump and a weekly recurrance for some job.

Yeah the weekly recurrence is for a horizontal autoscaler thingy--it basically polls constantly and the weekly recurrence being added makes it behave differently every weekend, instead of just one weekend a while ago. Version bump just references that change in Argo

PR #645, opened 2021-12-01
[IA-2938] Add Azure secrets to crljanitor chart

rtitle:
This updates the Janitor chart to read a new vault secret `secret/dsde/terra/azure/common/managed-app-publisher` and set it in the environment.

The secret path is set statically in `values` for now, but I expect it will need to be passed via `terra-helmfile` once we can Terraform different Azure environments (see https://broadworkbench.atlassian.net/browse/DDO-1772).

See related Janitor PR: https://github.com/DataBiosphere/terra-resource-janitor/pull/120

### DIFF
--- a/charts/crljanitor/templates/deployment.yaml
+++ b/charts/crljanitor/templates/deployment.yaml
@@ -112,6 +112,21 @@ spec:
             secretKeyRef:
               name: crl-janitor-admin-user-list
               key: admin-users
+        - name: AZURE_MANAGED_APP_CLIENT_ID
+          valueFrom:
+            secretKeyRef:
+              name: crl-janitor-azure-managed-app-creds
+              key: client-id
+        - name: AZURE_MANAGED_APP_CLIENT_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: crl-janitor-azure-managed-app-creds
+              key: client-secret
+        - name: AZURE_MANAGED_APP_TENANT_ID
+          valueFrom:
+            secretKeyRef:
+              name: crl-janitor-azure-managed-app-creds
+              key: tenant-id
         # Environment variables for OpenCensus resource auto detection.
         - name: CONTAINER_NAME
           valueFrom:
--- a/charts/crljanitor/values.yaml
+++ b/charts/crljanitor/values.yaml
@@ -39,7 +39,7 @@ vault:
   adminUserFilePath: config/terra/crl-janitor/common/iam
   # vault.azureManagedAppPublisherPath -- (string) Vault path for Azure managed app publisher credentials.
   # Uses the same path for all envs by default. Override if needed in helmfile repo.
-  azureManagedAppPublisherPath: secret/dsde/azure/common/managed-app-publisher
+  azureManagedAppPublisherPath: secret/dsde/terra/azure/common/managed-app-publisher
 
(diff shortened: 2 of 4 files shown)


charts/crljanitor/values.yaml
```diff
@@ -37,6 +37,9 @@ vault:
   # vault.adminUserFilePath -- (string) Vault path prefix for admin user list. Required if vault.enabled.
   # Use the same users as admin for all env by default. Override if needed in helmfile repo.
   adminUserFilePath: config/terra/crl-janitor/common/iam
+  # vault.azureManagedAppPublisherPath -- (string) Vault path for Azure managed app publisher credentials.
+  # Uses the same path for all envs by default. Override if needed in helmfile repo.
+  azureManagedAppPublisherPath: secret/dsde/terra/azure/common/managed-app-publisher
```
rtitle, line 42: Is this a good temp vault path? Open to other suggestions :)

mflinn-broad: works for me

charts/crljanitor/templates/deployment.yaml
```diff
@@ -112,6 +112,21 @@ spec:
...
+        - name: AZURE_MANAGED_APP_CLIENT_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: crl-janitor-azure-managed-app-creds
+              key: client-secret
+        - name: AZURE_MANAGED_APP_TENANT_ID
+          valueFrom:
+            secretKeyRef:
+              name: crl-janitor-azure-managed-app-creds
+              key: tenant-id
```
mflinn-broad, line 129:
```suggestion
       {{- if .Values.azureManagedApp.enabled }}
        - name: AZURE_MANAGED_APP_CLIENT_ID
          valueFrom:
            secretKeyRef:
              name: crl-janitor-azure-managed-app-creds
              key: client-id
        - name: AZURE_MANAGED_APP_CLIENT_SECRET
          valueFrom:
            secretKeyRef:
              name: crl-janitor-azure-managed-app-creds
              key: client-secret
        - name: AZURE_MANAGED_APP_TENANT_ID
          valueFrom:
            secretKeyRef:
              name: crl-janitor-azure-managed-app-creds
              key: tenant-id
        {{- end  }}
```

I'd reccomend wrapping this and the associated secretDefinition block this in a conditional and setting a default in `values.yaml` if we don't want this universally enabled in all environments yet. If we do want this enabled everywhere then feel free to disregard.

mflinn-broad (approved):
Lgtm with a possible suggestion!

charts/crljanitor/templates/deployment.yaml
rtitle: Good call.. `crljanitor` is a bit weird because there is no "live" environment -- the "production" instance is the `tools` personal environment: https://github.com/broadinstitute/terra-helmfile/blob/f01ea9932d1c400c36fb455896aa648d55a808c5/environments/personal/tools.yaml

But maybe it's a good idea to only enable Azure in the `tools` env..

mflinn-broad: Yeah my main concern here is potentially breaking any personal environments that may be using janitor. Although I'm not sure how much use those actually see these days anyways.

PR #647, opened 2021-12-01
[DDO-1729] Have Prometheus scrape Buffer

jack-r-warren:
ServiceMonitor disabled by default so as to not scrape from the personal environments


mflinn-broad (approved):
lgtm, thanks!

PR #649, opened 2021-12-03
[DDO-1579] Out with .Values.name, in with .Values.global.name

jack-r-warren:
- Added a note in the top-level readme about this being a universal standard instead of just a common pattern
- Moved every top-level `name:` in a Values file to be under `global:` (I did this manually to avoid regex mangling)
- Added a comment to every `global.name:` field
- Removed `liquibase-migration`'s duplicated name property in Buffer and WSM (would be unreferenced after this change)
- Regenerated all Chart-level readmes
- Find/replace `.Values.name` with `.Values.global.name` everywhere

For review:
1. Check out the README.md, it is top of the Files changed
2. You're welcome to spot check all the other changes but its like 244 files... the helmfile PR comments will handle telling us if this nukes or changes something weird

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,3 +31,11 @@ brew install norwoodj/tap/helm-docs
 cd charts/[changed chart]
 helm-docs
 ```
+
+## Standards
+
+We have a lot of patterns for charts that we tend to follow, but the standards here are meant to be universal.
+
+This list may grow over time; for a standard to be added here it must already be universal (in other words, this section is a reference for "what is," not a recommendation for "what should be").
+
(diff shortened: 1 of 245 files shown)


jack-r-warren:
Not WIP but putting "don't merge" here because automerge will vomit errors into devops channels if we just press the big green button here

choover-broad (approved):

jack-r-warren:
Merging now because https://github.com/broadinstitute/terra-helmfile/pull/2049 addresses the above comment

PR #651, opened 2021-12-06
DDO-1591 ArgoCD image bump

choover-broad:
Bump ArgoCD image for DDO-1591


choover-broad:
Merging now; will get retroactive approval since I'm doing off-hours maintenance

